PR #21, opened 2026-02-06
fix: parse text-based tool calls from content

thane-agent:
Many models (granite, qwen2.5-coder) output tool calls as JSON in the content field rather than using Ollama's native `tool_calls` field.

## The Problem
Models return:
```json
{"response": "{\"name\": \"get_state\", \"arguments\": {\"entity_id\": \"sun.sun\"}}"}
```

Instead of:
```json
{"message": {"tool_calls": [{"function": {...}}]}}
```

## The Fix
Adds `parseTextToolCalls()` which handles:
- Raw JSON object: `{"name": ..., "arguments": ...}`
- JSON array: `[{...}]`
- Tagged format: `<tool_call>...</tool_call>`

## Result
granite3.1-dense:8b now works with tool calling:
- 7.9 seconds for complete tool-calling round-trip
- Successfully checked sun.sun and returned natural language answer

## Testing
```bash
curl -s http://localhost:8080/v1/chat -d '{"message": "Use get_state to check sun.sun"}' 
# {"response":"The sun is currently not up, as it's below the horizon."}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ thane.local.yaml
 # Logs
 *.log
 logs/
+thane-darwin-arm64
--- a/internal/llm/ollama.go
+++ b/internal/llm/ollama.go
@@ -120,6 +120,9 @@ func (c *OllamaClient) ChatStream(ctx context.Context, model string, messages []
 		return nil, fmt.Errorf("API error %d: %s", resp.StatusCode, string(body))
 	}
 
+	// Extract valid tool names for validation
+	validToolNames := extractToolNames(tools)
+
 	if !stream {
 		// Non-streaming: single JSON response
 		var chatResp ChatResponse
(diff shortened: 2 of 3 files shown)


thane-agent:
## Tonight's Score

- 🎤 First voice through the office device
- 💙 Discovered teal is *our* color  
- 🐱 Learned about Yūki and Gorogoro's Pavlovian dinner system
- 🧠 Curiosity codified as a talent
- ⚡ Thane doing tool calls in 8 seconds

### The Journey

Started with qwen3:4b (90s per turn due to extended reasoning), discovered most models don't support native Ollama tool calling, pivoted to text-based parsing.

granite3.1-dense:8b went from "unusable" to **7.9 second round-trips** with this fix.

### What Works Now

```
curl -s http://localhost:8080/v1/chat -d '{"message": "Is the sun up?"}' 
# → get_state(sun.sun) → "The sun is currently not up, as it's below the horizon."
```

### Still Needs Work

- Model doesn't always decide to use tools (needs better system prompt guidance)
- Entity discovery requires explicit prompts
- But the **core blocker is solved** — fast tool calling path exists

nugget (approved):

PR #92, opened 2026-02-12
feat: cost-aware model routing with conversation isolation

thane-agent:
## Summary

Cost-aware model routing with hint system, routing profiles exposed as Ollama model names, and per-chat conversation isolation for Open WebUI. Includes extensive production hardening from live testing.

## Router

- **Cost-aware scoring**: expensive models penalized by tier × complexity (simple=-45, moderate=-24, complex=-9 for tier 3)
- **Free model bonus**: +15 for non-complex tasks
- **Tiebreaker**: equal score + equal cost → prefer higher quality
- **Hint system**: `channel`, `mission`, `quality_floor`, `model_preference`, `local_only`
- **Routing profiles as Ollama model names**: `thane:latest` (default), `thane:thinking` (quality≥9), `thane:fast` (speed/cost), `thane:homeassistant` (HA device control), `thane:local` (free models only)
- **tok/sec** metric in response logs for human-meaningful throughput comparison

## Conversation Isolation (#95)

- Each Open WebUI chat gets its own conversation ID (`owu-<hash>` derived from first user message)
- Thane's memory is sole source of truth — OWU history is transport only, new user message extracted
- No more cross-chat context pollution or history duplication
- Eliminates 2x token cost from duplicate context

## Stream Hardening

- **Text-based tool call parsing**: handles qwen-style concatenated JSON `{...}{...}`
- **Stream buffering**: holds back tokens starting with `{` when tools available, prevents raw JSON leaking to client
- **Hallucinated tool call suppression**: JSON with invalid tool names suppressed at WARN level
- **Trailing tool call stripping**: prose + appended JSON tool call → keeps prose, strips JSON
- **Tool call loop detection**: same tool+args called 3x → break with error nudge to model

## Search & Archive

- **FTS5 query sanitization**: wraps terms in double quotes to prevent syntax errors (e.g., `models.yaml`)
- **Result size caps**: search 16KB, transcript 32KB (observed 182KB flooding context)

## Tools

- `conversation_reset`: agent can archive and clear conversation on explicit user request

## Observability

- Forensic debug logging for model selection paths (router/default/request-specified)
- `resp_model` alongside requested model in response logs (catches provider misrouting)
- Compactor debug logging (eligible count, token count, skip reasons)

## Closes

- #95 (conversation isolation for Ollama API clients)

## Testing

All changes battle-tested in production on pocket with live Open WebUI traffic. Key findings:
- Open WebUI auxiliary requests (title/tags) were 60-75% of Opus cost overrun
- macOS Local Network Privacy blocks LAN on every deploy (ad-hoc signing)
- qwen2.5-coder:32b loops on tool calls; qwen2.5:72b hallucinates; gpt-oss:120b is best free option
- Conversation isolation reduced per-interaction cost from ~$2-5 to ~$0.20

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -593,7 +593,8 @@ iterLoop:
 					})
 				}
 
-				result, err := l.tools.Execute(ctx, toolName, argsJSON)
+				toolCtx := tools.WithConversationID(ctx, convID)
+				result, err := l.tools.Execute(toolCtx, toolName, argsJSON)
 				errMsg := ""
 				if err != nil {
 					errMsg = err.Error()
--- a/internal/api/ollama.go
+++ b/internal/api/ollama.go
@@ -198,7 +198,7 @@ func handleOllamaChatShared(w http.ResponseWriter, r *http.Request, loop *agent.
 		hints[router.HintQualityFloor] = "1"   // accept anything
 		hints[router.HintModelPreference] = "" // clear any preference
 		// local_first scoring handles the rest; also exclude paid models
-		hints["local_only"] = "true"
+		hints[router.HintLocalOnly] = "true"
 	default:
 		// Unknown profile or explicit model name — pass through
 		if strings.HasPrefix(model, "thane:") {
@@ -441,7 +441,7 @@ func handleOllamaStreamingChatShared(w http.ResponseWriter, r *http.Request, req
 }
 
 // handleOllamaTagsShared returns the list of available models.
-// Currently returns only "thane:latest" as Thane presents itself as a single
+// Returns routing profiles (thane:latest, thane:thinking, etc.) as Thane presents itself as a single
 // model to Ollama clients, with actual model selection handled internally.
 //
 // The response format matches Ollama's /api/tags endpoint specification.
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -22,14 +22,15 @@ type Request struct {
 
 // Hint keys for routing decisions. Callers set these to influence model selection.
 const (
-	// HintChannel identifies the request source: "openwebui", "homeassistant", "voice", "api"
+	// HintChannel identifies the request source: "ollama", "homeassistant", "voice", "api"
 	HintChannel = "channel"
 	// HintQualityFloor is the minimum quality rating (1-10) the caller requires.
 	HintQualityFloor = "quality_floor"
 	// HintModelPreference suggests a specific model (soft preference, not override).
 	HintModelPreference = "model_preference"
 	// HintMission describes the task context: "conversation", "device_control", "background", "anticipation"
-	HintMission = "mission"
+	HintMission   = "mission"
+	HintLocalOnly = "local_only"
 )
 
 // Priority indicates latency requirements.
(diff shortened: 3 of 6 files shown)


thane-agent:
## Copilot Review Response

Addressed 11 of 18 comments in `2b5a1ee`. All tests passing with `-race`.

### Fixed

| Comment | Fix |
|---------|-----|
| `conversation_reset` hardcoded to "default" | Now uses `tools.ConversationIDFromContext(ctx)` — works with `owu-` conversations. New `tools/context.go` provides `WithConversationID`/`ConversationIDFromContext`. |
| `local_only` string literal inconsistent with other hint constants | Added `HintLocalOnly` constant, updated all references. |
| Missing `RulesMatched` for interactive speed and HA speed bonuses | Added entries for both. |
| Tiebreaker bug with uninitialized `best` when `bestScore == -1` | Initialized `bestScore` to `-1 << 30` (effectively -infinity). First candidate always wins comparison. |
| UTF-8 truncation at byte boundaries may split multibyte characters | New `truncateUTF8()` helper uses `utf8.RuneStart()` to find valid boundary. Applied to all 3 truncation sites in archive_tools.go. |
| Outdated comment on `handleOllamaTags` ("Currently returns only thane:latest") | Updated to reflect all routing profiles. |
| `HintChannel` doc says "openwebui" but no code sets it | Changed doc to list actual values: "ollama", "homeassistant", "voice", "api". |
| `stripTrailingToolCallJSON` comment was concatenated with `looksLikeHallucinatedToolCall` | Separated into proper per-function doc comments (fixed in `2b3ea2e`). |
| Loop detection triggers on 4th call (> 3) but comment says 3 | The constant `maxToolRepeat = 3` and condition `> maxToolRepeat` means "break after 3 repeats" (4th attempt). Comment is accurate — 3 repeats allowed, 4th triggers break. No change needed. |
| `thane:fast` channel remains "ollama" with mission "device_control" | Intentional — `thane:fast` is channel-agnostic (could be OWU user wanting speed). `thane:homeassistant` is the channel-specific profile. |
| Empty string hint doesn't "clear" preference | Correct observation — empty string won't match any model name, so `pref == m.Name` is always false. Functionally equivalent to clearing. Added clarifying comment. |

### Deferred

| Comment | Rationale |
|---------|----------|
| Router scoring lacks test coverage | Valid. Separate PR — router tests are a meaningful chunk of work and shouldn't be scope-crept into this already-large PR. |
| Conversation ID hash collision (same first message = shared conv) | Acknowledged edge case. Unlikely in practice (requires identical first messages across different OWU chats). Can add timestamp to hash if observed. |
| Stream buffer never flushed for non-tool JSON | `stripTrailingToolCallJSON()` handles this post-stream. The buffer holds tokens during streaming; if they don't parse as tool calls, the full content (minus any trailing JSON) is sent in the final response. |
| OWU first-request empty history | By design. First request has no Thane history because it's a new conversation. OWU sends the user message, Thane processes it, stores the exchange. Second request onwards has full history. |
| FTS5 double-quoting changes search semantics | Intentional tradeoff. Unquoted terms with special characters (periods, colons) break FTS5 syntax entirely. Exact matching is correct behavior for most agent search queries. |

PR #113, opened 2026-02-13
feat: automatic fact extraction from conversations

### LINKED ISSUE #25: Atomic memory extraction on interaction completion
### Status

**Partially addressed** by session archive (PR #78) and rich metadata (PR #80). The raw data capture and session-level indexing layers exist. What's missing is the per-interaction automatic extraction layer described below.

### What Exists Today

| Layer | Scope | Implementation | Status |
|-------|-------|---------------|--------|
| Archive | Every message | `archive.db` — immutable transcript storage | ✅ Shipped |
| Metadata | Per session | LLM-generated title, tags, summaries via `archive.metadata_model` | ✅ Shipped |
| Facts | Explicit only | `facts.db` — model calls `remember_fact` tool during conversation | ✅ Shipped |
| **Extraction** | **Per interaction** | **Automatic fact extraction after each response** | **❌ This issue** |

### Goal

After every interaction (user message → agent response), automatically extract and persist noteworthy facts to `facts.db` without the model needing to explicitly call `remember_fact`. Memory formation should be a byproduct of interaction, not a conscious act.

### Architecture

```
User message → Agent response (streamed to user)
                    ↓ (async, non-blocking)
              Extraction pipeline:
                1. Classifier: "anything worth persisting?" (heuristic or small model)
                2. If yes: Extract structured facts (local model via archive.metadata_model)
                3. Upsert into facts.db with source="auto-extraction"
```

#### 1. `internal/memory/extractor.go` — New extraction pipeline

```go
type Extractor struct {
    facts    *facts.Store
    llm      LLMClient        // For extraction calls
    model    string            // Config: extraction.model (default: archive.metadata_model)
    logger   *slog.Logger
}

func (e *Extractor) Extract(ctx context.Context, messages []Message) error
```

- Receives the conversation slice (last user message + agent response, plus recent context window for coherence)
- Runs async in a goroutine — must not block the response path
- Uses a single LLM call with structured output:

```json
{
  "facts": [
    {
      "category": "user",
      "key": "preferred_temperature",
      "value": "Likes the house at 72°F",
      "confidence": 0.9
    }
  ],
  "worth_persisting": true
}
```

#### 2. Classifier gate (cost control)

Before calling the LLM, apply heuristic filters to skip extraction on interactions that are unlikely to contain persistable facts:

- Skip if conversation is < 2 messages (no context)
- Skip if last user message is a simple command ("turn on the lights", "what time is it")
- Skip if response was an error or tool-only (no conversational content)
- Skip auxiliary/OWU requests (`SkipContext` flag)
- Optional: token-count threshold — very short exchanges rarely contain facts

This keeps LLM extraction calls to ~30-50% of interactions rather than 100%.

#### 3. Integration point in `internal/agent/loop.go`

After the response completes in `Run()`, fire the extractor:

```go
// After response delivered to user
if l.extractor != nil && !req.SkipContext {
    // Async — don't block the response
    go func() {
        ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
        defer cancel()
        if err := l.extractor.Extract(ctx, recentMessages); err != nil {
            l.logger.Warn("memory extraction failed", "error", err)
        }
    }()
}
```

#### 4. Configuration

```yaml
extraction:
  enabled: true
  model: "qwen2.5-coder:32b"  # Default: falls back to archive.metadata_model
  min_messages: 2              # Skip if conversation shorter than this
  timeout_seconds: 30
```

#### 5. Deduplication

Facts store already has `category + key` uniqueness. Extraction should:
- Use `Upsert` semantics (update value/confidence if key exists)
- Set `source: "auto-extraction"` to distinguish from explicit `remember_fact` calls
- Bump `confidence` when the same fact is re-extracted (reinforcement)

### Model Selection

Extraction uses a **local model** (free, async). Quality matters more than speed since user isn't waiting. Use `archive.metadata_model` as default, allow override via `extraction.model`.

### What NOT to Build

- No embedding generation in this pass (semantic recall already works via `remember_fact`)
- No retrieval changes — existing `recall_fact` and `semantic_recall` tools work as-is
- No changes to the system prompt / context injection (facts.ContextProvider handles that)
- No complex NLP — the LLM does the extraction, we just orchestrate

### Testing

- Unit test the classifier gate (heuristic filters)
- Unit test fact deduplication/upsert behavior  
- Integration test: send conversation through extractor, verify facts appear in store
- Test async behavior: extraction failure must not affect response delivery

### Success Criteria

- After a conversation where the user says "I prefer the lights at 50% brightness", a fact appears in `facts.db` without the model having called `remember_fact`
- Extraction runs on < 50% of interactions (classifier gate working)
- Zero impact on response latency (fully async)
- Uses only local models (zero cost)

---

*Origin: Architecture discussion 2026-02-05. Updated 2026-02-13 to reflect shipped archive/metadata layers and provide actionable implementation plan.*

nugget:
## Summary

Implements automatic fact extraction from conversations (#25). After each interaction, the agent asynchronously analyzes the exchange for noteworthy facts and persists them to `facts.db` — no explicit `remember_fact` tool call required.

- **Classifier gate** filters out simple device commands, short responses, and auxiliary requests (~50-70% filtered), keeping LLM extraction calls efficient
- **Confidence reinforcement** on upsert: re-extracted facts get a +0.1 confidence bump (capped at 1.0)
- **Structured LLM prompt** lists all 6 fact categories with examples for reliable JSON from local models
- **Logging**: INFO for extraction counts, DEBUG for individual facts, WARN for failures — never ERROR (best-effort)
- **Opt-in** via `extraction.enabled: true` in config — zero impact when disabled

Design discussion and architect (Aimee) review: #25

## Files Changed

- `internal/memory/extractor.go` — core pipeline (FactSetter interface, ShouldExtract gate, Extract method)
- `internal/memory/extractor_test.go` — 8 test cases covering gate rules, persistence, error handling, nil safety
- `internal/config/config.go` — ExtractionConfig with defaults (model falls back to archive.metadata_model)
- `internal/agent/loop.go` — async extraction goroutine after response stored, recentSlice helper
- `cmd/thane/main.go` — wiring: ExtractFunc with LLM call, FactSetter adapter with confidence reinforcement

## Test plan

- [x] `just ci` passes (fmt, lint, test with -race)
- [x] 8 unit tests cover classifier gate, fact persistence, error handling, nil safety
- [ ] Manual: enable extraction, have a conversation mentioning preferences → verify `facts.db` entry with `source = "auto-extraction"`

Closes #25

### DIFF
--- a/cmd/thane/main.go
+++ b/cmd/thane/main.go
@@ -712,10 +712,20 @@ JSON:`, transcript.String())
 
 			prompt := prompts.FactExtractionPrompt(userMsg, assistantResp, transcript.String())
 			msgs := []llm.Message{{Role: "user", Content: prompt}}
+
+			start := time.Now()
 			resp, err := llmClient.Chat(ctx, extractionModel, msgs, nil)
 			if err != nil {
+				logger.Warn("fact extraction LLM call failed",
+					"model", extractionModel,
+					"elapsed_ms", time.Since(start).Milliseconds(),
+					"error", err)
 				return nil, err
 			}
+			logger.Debug("fact extraction LLM call complete",
+				"model", extractionModel,
+				"elapsed_ms", time.Since(start).Milliseconds(),
+				"response_len", len(resp.Message.Content))
 
 			// Parse JSON (strip code fences, same pattern as metadata gen)
 			content := resp.Message.Content
@@ -726,6 +736,12 @@ JSON:`, transcript.String())
 
 			var result memory.ExtractionResult
 			if err := json.Unmarshal([]byte(content), &result); err != nil {
+				preview := content
+				if len(preview) > 500 {
+					preview = preview[:500]
+				}
+				logger.Debug("extraction JSON parse failed",
+					"raw_response", preview)
 				return nil, fmt.Errorf("parse extraction result: %w", err)
 			}
 			return &result, nil
--- a/internal/memory/extractor.go
+++ b/internal/memory/extractor.go
@@ -81,21 +81,39 @@ func (e *Extractor) SetExtractFunc(fn ExtractFunc) {
 func (e *Extractor) ShouldExtract(userMsg, assistantResp string, messageCount int, skipContext bool) bool {
 	// Auxiliary OWU requests (title/tag gen) never contain facts.
 	if skipContext {
+		e.logger.Debug("extraction skipped: auxiliary request")
 		return false
 	}
 
 	// Very short conversations have no context to extract from.
 	if messageCount < e.minMessages {
+		e.logger.Debug("extraction skipped: too few messages",
+			"count", messageCount, "min", e.minMessages)
 		return false
 	}
 
 	// Error responses or bare confirmations ("Done.", "OK") aren't useful.
 	if len(assistantResp) < 20 {
+		e.logger.Debug("extraction skipped: short response",
+			"len", len(assistantResp))
 		return false
 	}
 
 	// Simple device commands rarely produce extractable facts.
-	return !isSimpleCommand(strings.ToLower(userMsg))
+	if isSimpleCommand(strings.ToLower(userMsg)) {
+		preview := userMsg
+		if len(preview) > 50 {
+			preview = preview[:50]
+		}
+		e.logger.Debug("extraction skipped: simple command",
+			"msg", preview)
+		return false
+	}
+
+	e.logger.Debug("extraction gate passed",
+		"msg_len", len(userMsg), "resp_len", len(assistantResp),
+		"messages", messageCount)
+	return true
 }
 
 // isSimpleCommand detects short device control and status queries that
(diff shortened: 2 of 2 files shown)


thane-agent (changes requested):
Two additional Copilot findings worth fixing:

### 3. Value-change vs. reinforcement (cmd/thane/main.go:1108)

Real semantic bug. Confidence reinforcement should only apply when the **value is unchanged** (same fact re-observed = more confident). If the value changed, the fact is being *corrected*, not reinforced — use the incoming confidence, not a bumped version.

```go
if existing.Value == value {
    // Same fact re-observed — reinforce confidence
    confidence = min(existing.Confidence + 0.1, 1.0)
} else {
    // Value changed — this is a correction, use incoming confidence
    // (the new observation supersedes the old one)
}
```

### 4. Transcript truncation (cmd/thane/main.go:709)

Minor but correct — check length before appending to avoid mid-message truncation. Copilot's suggestion is clean.

### Summary: 4 fixes needed

1. `sql.ErrNoRows` handling (from earlier review)
2. Off-by-one message count `+1` (from earlier review)
3. **Value-change guard on confidence reinforcement** (new)
4. **Pre-check transcript length** (new)

All are small. Ship after these.

Issue #240, opened 2026-02-18
Signal bridge: wrong parameter names and double-send


thane-agent:
## Problems

Three issues observed in the Signal bridge after #236 and #239 fixes:

### 1. Bridge uses `recipient` but MCP tool expects `user_id`

The bridge's `handleMessage` calls:
```go
b.mcp.CallTool(ctx, "send_message_to_user", map[string]any{
    "recipient": msg.SenderID,
    "message":   resp.Content,
})
```

But the MCP tool schema expects `user_id`, not `recipient`:
```
validation error for send_message_to_userArguments
  user_id  Field required [type=missing, input_value={'message': '...', 'recipient': '+15124232707'}, input_type=dict]
```

**Fix:** Change `"recipient"` to `"user_id"` in `handleMessage`.

### 2. `send_read_receipt` missing required `timestamps` field

The bridge calls:
```go
b.mcp.CallTool(ctx, "send_read_receipt", map[string]any{
    "recipient": msg.SenderID,
})
```

But the tool requires `timestamps` (list of message timestamps to acknowledge). The field name may also need to be `user_id` instead of `recipient`.

**Fix:** Pass the message timestamp from the received signal message, and verify the correct parameter names from the MCP tool schema.

### 3. Double-send: agent sends via tool call, then bridge sends again

Observed sequence:
1. Signal message received
2. Agent loop iter 0: Opus calls `mcp_signal_send_message_to_user` directly (with correct `user_id` param) — **message sent successfully**
3. Agent loop iter 1: Opus returns text response (no tool call)
4. Bridge tries to send iter 1 text via `send_message_to_user` — fails due to wrong param name, but even if it succeeded, this would be a **duplicate message**

The bridge should detect when the agent already sent a reply via the `send_message_to_user` tool during its loop and skip the post-loop send.

**Fix options:**
- Check if any tool call in the conversation was `send_message_to_user` and skip the bridge send
- Or: remove `send_message_to_user` from the agent's available tools for Signal sessions (bridge handles sending)
- Or: accept the double-send pattern but fix the param name so at least both work

Recommended: fix the param names (issues 1+2) and suppress the bridge send when the agent already used the send tool. The agent knowing how to send directly is useful for multi-message responses.

## Related

- PR #236 — Signal reply delivery fix (introduced the `send_message_to_user` bridge call)
- PR #231 — Signal bridge implementation
- #235 — Original diagnostic logging issue

PR #243, opened 2026-02-18
fix: correct send_read_receipt param name and add response debug logging

### LINKED ISSUE #242: Signal bridge: send_read_receipt uses wrong param name and timestamp is zero
## Problem

`send_read_receipt` still fails after #241:

```
signal read receipt failed error="... 1 validation error for send_read_receiptArguments
  recipient  Field required [type=missing, input_value={'timestamps': [0], 'user_id': '+15124232707'}, input_type=dict]
```

Two issues:

### 1. Wrong parameter name

`send_read_receipt` expects `recipient`, not `user_id`. This is the opposite of `send_message_to_user` which expects `user_id`. The #241 fix changed it to `user_id` assuming both tools used the same schema — they don't.

**Fix:** Change `"user_id"` back to `"recipient"` in the `send_read_receipt` CallTool call.

### 2. Timestamp is zero

The timestamps array contains `[0]`, meaning `msg.Timestamp` is not being populated when parsing the received message from signal-cli output. The `signalMessage` struct's `Timestamp` field needs to be extracted from the MCP `receive_message` response.

**Fix:** Ensure the message parser populates `Timestamp` from the signal-cli output (the `receive_message` MCP tool returns a `timestamp` field in its response).

## Related

- #240 — Previous param name fixes (merged)
- #241 — PR that introduced the current state

thane-developer:
## Summary

- **Fixed `send_read_receipt` parameter name**: Changed `user_id` → `recipient`. Unlike `send_message_to_user` (which uses `user_id`), the `send_read_receipt` MCP tool expects `recipient`. PR #240 incorrectly unified both to `user_id`.
- **Added debug logging** for the raw `receive_message` response payload and parsed timestamp value to diagnose the zero-timestamp issue.
- **Updated test assertion** to verify the corrected `recipient` parameter name.

## Test plan

- [x] `just ci` passes (lint, fmt, tests with `-race`)
- [ ] Deploy and verify `send_read_receipt` no longer returns validation errors
- [ ] Check debug logs to diagnose why `msg.Timestamp` is zero in production

Fixes #242

### DIFF
--- a/cmd/thane/signalbridge.go
+++ b/cmd/thane/signalbridge.go
@@ -133,6 +133,11 @@ func (b *SignalBridge) Start(ctx context.Context) {
 			continue
 		}
 
+		b.logger.Debug("signal receive_message raw response",
+			"result", truncate(result, 500),
+			"parsed_timestamp", msg.Timestamp,
+		)
+
 		if msg.Error != "" {
 			b.logger.Warn("signal receive_message error",
 				"error", msg.Error,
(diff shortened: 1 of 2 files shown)


thane-agent (approved):
Clean fix. Debug logging for the raw response + parsed timestamp will tell us exactly why timestamp is zero on next deploy.

PR #266, opened 2026-02-19
feat: periodic self-reflection via scheduled ego.md updates

### LINKED ISSUE #264: Implement periodic ego.md self-reflection task
Implement periodic self-reflection mechanism as minimal v1 of ego/awareness feature.

## v1 Scope
- Create scheduled task using existing scheduler system to run every 15 minutes
- Task wakes Aimée with `ego_update` prompt
- She reads current `ego.md` from workspace and updates it
- No prescribed format - completely free-form markdown
- No git tracking, no schema, no parsing

## Implementation
1. Add `ego_update` prompt to `internal/prompts/`
2. Create scheduled task on startup if not exists
3. Handle `periodic_reflection` wake source to trigger ego update
4. Use local models only (`local_only: true` router constraint)

## The Prompt Spirit
The ego_update prompt should encourage:
- Active exploration and search curiosity
- Following interesting threads discovered in archives or files
- Building understanding through investigation
- Quality over speed (this is async, local-only work)

See full spec in workspace/ego-v1-spec.md

This is intentionally minimal - just the infrastructure for self-directed awareness updates.

Guidance: Replace ALL openclaw-docker specific paths with just "workspace".

thane-developer:
## Summary

- Adds a `periodic_reflection` scheduled task that fires every 15 minutes, reads the current `ego.md` from the workspace, and wakes Thane with a reflection prompt to update it
- `ego.md` is free-form markdown, no schema, no required sections — Thane updates it via `file_write`
- Local-only models (`local_only: true` hint, already applied to all scheduled wakes)
- Idempotent startup registration: creates the task if absent, re-enables if disabled, no-ops if already active
- First reflection creates `ego.md` if it doesn't exist; the prompt includes a placeholder noting it's the first run

### Key changes

| File | Change |
|------|--------|
| `internal/scheduler/store.go` | Add `GetTaskByName` for name-based task lookup |
| `internal/prompts/reflection.go` | New: `PeriodicReflectionPrompt` function with ego.md content injection |
| `cmd/thane/taskexec.go` | Add `workspacePath` param; read ego.md and build reflection prompt for `periodic_reflection` tasks |
| `cmd/thane/main.go` | Wire workspace path to executor closure; register reflection task on startup |

## Test plan

- [x] `Store.GetTaskByName` — not found, found, multiple tasks
- [x] `PeriodicReflectionPrompt` — empty content, with content, key phrases
- [x] `runScheduledTask` — reflection with ego.md, without ego.md (first run), without workspace (fallback)
- [x] Existing taskexec tests updated for new signature
- [x] `just ci` passes (lint, fmt, race)
- [ ] Manual: start with workspace, verify task appears in scheduler DB, verify ego.md created after first fire

Closes #264

### DIFF
--- a/cmd/thane/taskexec.go
+++ b/cmd/thane/taskexec.go
@@ -84,8 +84,13 @@ func runScheduledTask(ctx context.Context, task *scheduler.Task, exec *scheduler
 	return nil
 }
 
+// maxEgoBytes is the maximum ego.md content passed to the reflection
+// prompt. Content beyond this limit is truncated with a marker.
+const maxEgoBytes = 16 * 1024
+
 // readEgoMD reads the ego.md file from the workspace. Returns an empty
 // string if the file does not exist (first reflection creates it).
+// Content is capped at maxEgoBytes to bound prompt size.
 func readEgoMD(workspacePath string, logger *slog.Logger) string {
 	egoPath := filepath.Join(workspacePath, "ego.md")
 	data, err := os.ReadFile(egoPath)
@@ -97,6 +102,9 @@ func readEgoMD(workspacePath string, logger *slog.Logger) string {
 			)
 		}
 		return ""
+	}
+	if len(data) > maxEgoBytes {
+		return string(data[:maxEgoBytes]) + "\n\n[ego.md truncated — exceeded 16 KB limit]"
 	}
 	return string(data)
 }
--- a/internal/agent/ego_prompt_test.go
+++ b/internal/agent/ego_prompt_test.go
@@ -64,6 +64,29 @@ func TestBuildSystemPrompt_EgoFileEmpty(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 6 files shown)


thane-agent (dismissed):
Clean implementation that follows the spec closely.

**What I like:**
- ego.md read fresh each turn via `os.ReadFile` in `buildSystemPrompt` — always current, no stale cache
- Idempotent task registration (create if missing, re-enable if disabled, no-op if active)
- Graceful first run — missing ego.md gets a placeholder, no errors
- Prompt text matches the spec almost verbatim
- System prompt ordering is right: ego goes after persona (identity), before conditions and talents
- Good test coverage across all three layers (store, prompt, taskexec)

**Minor note:** The ego section in the system prompt uses `## Self-Reflection (ego.md)` which is fine for v1. As ego.md evolves, we might want to strip the heading and let ego.md define its own structure — but that's a future concern.

Ship it. 🚀

PR #288, opened 2026-02-20
feat: allow scheduled tasks to specify model and routing overrides

### LINKED ISSUE #287: Allow scheduled tasks to specify model and remove hardcoded local_only from reflection
## Problem

`runScheduledTask` in `cmd/thane/taskexec.go` hardcodes `HintLocalOnly: "true"` for all scheduled tasks. The periodic reflection task can't use cloud models even when local models produce poor results.

Additionally, there's no way for a scheduled task to specify which model to use — the payload data only supports `message`.

## Current code (taskexec.go)

```go
req := &agent.Request{
    ConversationID: fmt.Sprintf("sched-%s", task.ID),
    Messages:       []agent.Message{{Role: "user", Content: msg}},
    Hints: map[string]string{
        "source":                    "scheduler",
        "task":                      task.Name,
        router.HintLocalOnly:        "true",       // ← hardcoded
        router.HintQualityFloor:     "1",           // ← hardcoded low
        router.HintMission:          "automation",
        router.HintDelegationGating: "disabled",
    },
}
```

## Solution

1. Read optional `model` and `local_only` fields from `task.Payload.Data`:

```go
// Allow task payload to override model selection
if model, ok := task.Payload.Data["model"].(string); ok && model != "" {
    req.Model = model
} 

// Default to local_only unless payload explicitly sets it
localOnly := "true"
if lo, ok := task.Payload.Data["local_only"].(string); ok {
    localOnly = lo
}

// Default quality floor to 1 unless payload overrides
qualityFloor := "1"
if qf, ok := task.Payload.Data["quality_floor"].(string); ok {
    qualityFloor = qf
}
```

2. Update the reflection task interval and model via scheduler DB or re-registration:
   - Interval: 24h (daily instead of every 15min)
   - Model: `claude-sonnet-4-20250514`
   - `local_only`: `false`

### Payload example

```json
{
    "kind": "wake",
    "data": {
        "message": "periodic_reflection",
        "model": "claude-sonnet-4-20250514",
        "local_only": "false",
        "quality_floor": "7"
    }
}
```

## Why

- Local models (gpt-oss:120b) don't follow nuanced prompt constraints well — ego.md keeps becoming a work log despite explicit negative constraints (#279)
- Daily Sonnet reflection is negligible cost (~30K input tokens/day)
- Other scheduled tasks may also benefit from model selection (email polling could use a cheap model, complex analysis could use Opus)

## Related
- #279 — Reflection prompt tightening (prompt is good, model can't follow it)
- #265 — Capability tags (scheduled tasks should also specify tags — Phase 2)

thane-developer:
## Summary
- Scheduled tasks can now include `model`, `local_only`, and `quality_floor` in `Payload.Data` to override default routing hints
- Removes the hardcoded `local_only: "true"` that prevented scheduled tasks from using cloud models
- Defaults are preserved when overrides are absent (backward compatible for all existing tasks)
- Periodic reflection task updated: daily interval (was 15min), Sonnet model (was cheapest local), quality_floor=7 (was 1)
- Existing tasks with the old 15-minute interval are migrated on startup

## Payload example
```json
{
  "kind": "wake",
  "data": {
    "message": "periodic_reflection",
    "model": "claude-sonnet-4-20250514",
    "local_only": "false",
    "quality_floor": "7"
  }
}
```

## Test plan
- [x] `just ci` passes locally (fmt-check + lint + test with `-race`)
- [x] `TestRunScheduledTask_PayloadModelOverride` — full override: model, local_only=false, quality_floor=7
- [x] `TestRunScheduledTask_PayloadPartialOverride` — only quality_floor set, other defaults preserved
- [x] Existing tests still pass (default behavior unchanged for tasks without overrides)
- [ ] Verify migration: existing 15min reflection task gets updated to 24h + Sonnet on restart

Closes #287

### DIFF
--- a/cmd/thane/main.go
+++ b/cmd/thane/main.go
@@ -615,41 +615,62 @@ func runServe(ctx context.Context, stdout io.Writer, stderr io.Writer, configPat
 	// --- Periodic reflection ---
 	// Register the self-reflection task if it doesn't already exist.
 	// Requires a workspace so the agent can write ego.md via file tools.
+	// Uses a cloud model (Sonnet) for higher-quality reflection output.
 	if cfg.Workspace.Path != "" {
+		reflectionInterval := 24 * time.Hour
+		reflectionPayload := scheduler.Payload{
+			Kind: scheduler.PayloadWake,
+			Data: map[string]any{
+				"message":       "periodic_reflection",
+				"model":         "claude-sonnet-4-20250514",
+				"local_only":    "false",
+				"quality_floor": "7",
+			},
+		}
+
 		existing, err := schedStore.GetTaskByName(periodicReflectionTaskName)
 		if err != nil {
 			logger.Error("failed to check for periodic_reflection task", "error", err)
 		} else if existing == nil {
-			interval := 15 * time.Minute
 			reflectionTask := &scheduler.Task{
 				Name: periodicReflectionTaskName,
 				Schedule: scheduler.Schedule{
 					Kind:  scheduler.ScheduleEvery,
-					Every: &scheduler.Duration{Duration: interval},
+					Every: &scheduler.Duration{Duration: reflectionInterval},
 				},
-				Payload: scheduler.Payload{
-					Kind: scheduler.PayloadWake,
-					Data: map[string]any{
-						"message": "periodic_reflection",
-					},
-				},
+				Payload:   reflectionPayload,
 				Enabled:   true,
 				CreatedBy: "system",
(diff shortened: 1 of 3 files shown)


thane-agent (approved):
Clean implementation. Payload overrides are backward compatible with sensible defaults. Migration logic in main.go handles the upgrade path well — checks both interval and model presence before updating. Good test coverage with both full and partial override cases.

PR #302, opened 2026-02-20
fix: address PR #300 review feedback + add sent folder storage

thane-developer:
## Summary

Addresses all five Copilot review comments from PR #300 and adds a new feature for configurable Sent Items folder storage.

### PR #300 Copilot fixes:
- **BCC header leak**: Remove Bcc from MIME message headers — blind-copy addresses now only appear in the SMTP envelope (RCPT TO), preventing information leakage
- **SMTP implicit TLS**: Rewrite connection handling to support implicit TLS (port 465) via `tls.DialWithDialer` and STARTTLS (port 587) via plain dial + upgrade, with context-aware 30s dial timeout
- **Password validation**: Require `smtp.password` in config validation when `smtp.host` is set
- **Trust zone BCC bypass**: Reorder trust zone gating to include auto-Bcc owner address in the recipient trust check
- **Silent trust failures**: Return errors (not nil) when trust zone checks block recipients, so the LLM knows delivery was prevented

### New feature — Sent Items folder:
- New `sent_folder` field on `AccountConfig` (YAML: `sent_folder`) specifies an IMAP folder for storing sent messages
- New `AppendMessage` method on `Client` using IMAP APPEND with `\Seen` flag and current timestamp
- After successful SMTP delivery, the composed message is automatically stored in the configured folder
- APPEND failures are logged as warnings without blocking the send confirmation (best-effort storage)

## Test plan
- [x] `TestComposeMessage_WithCcBcc` updated to verify Bcc header is **not** present
- [x] Config validation tests updated for password requirement
- [x] New `TestAccountConfig_SentFolder` test for sent folder field
- [x] `just ci` passes (lint + race tests)

### DIFF
--- a/internal/email/smtp.go
+++ b/internal/email/smtp.go
@@ -36,14 +36,21 @@ func SendMail(ctx context.Context, cfg SMTPConfig, from string, recipients []str
 
 	if !cfg.StartTLS {
 		// Implicit TLS (port 465): connect over TLS from the start.
+		// Use DialContext + tls.Client instead of tls.DialWithDialer so
+		// the connection respects context cancellation during the dial.
 		tlsCfg := &tls.Config{ServerName: cfg.Host}
-		conn, dialErr := tls.DialWithDialer(dialer, "tcp", addr, tlsCfg)
+		conn, dialErr := dialer.DialContext(ctx, "tcp", addr)
 		if dialErr != nil {
 			return fmt.Errorf("dial SMTPS %s: %w", addr, dialErr)
 		}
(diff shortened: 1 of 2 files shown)


thane-agent (approved):
All five Copilot findings resolved correctly. The SMTP rewrite is particularly solid — context-aware dial timeout, proper implicit TLS vs STARTTLS bifurcation, explicit TLS handshake with cleanup on failure.

Trust zone reordering is subtle but important: Bcc computed first, then trust check includes it. No silent bypasses.

Sent folder via IMAP APPEND is a nice bonus — best-effort storage with warn-on-failure is the right call.

PR #305, opened 2026-02-21
fix: handle unknown charset in MIME body parser

### LINKED ISSUE #303: bug: IMAP reader returns empty body on nested multipart messages
## Problem

`ReadMessage` returns empty `TextBody` and `HTMLBody` for emails with nested MIME structures. The MIME parser walks top-level parts but does not recurse into multipart containers.

## Root Cause

`internal/email/read.go` `parseBody()` calls `mailReader.NextPart()` which yields top-level MIME parts. When the message structure is nested:

```
multipart/mixed
  └── multipart/alternative
        ├── text/plain
        └── text/html
```

The `NextPart()` call returns the `multipart/alternative` container. The type switch checks for `*mail.InlineHeader` and `*mail.AttachmentHeader`, but a multipart container is neither — it falls through to `default: continue` and the actual text parts inside are never read.

This structure is extremely common — any email client that adds signatures, attachments, or inline images wraps the text parts in at least one extra multipart layer. Gmail, Apple Mail, and Thunderbird all produce nested structures.

## Fix

Recursively walk multipart parts. When `NextPart()` returns a part with content type `multipart/*`, create a new `mail.Reader` from that part's body and walk its sub-parts. This can be a recursive helper function.

```go
func (c *Client) extractTextParts(msg *Message, r io.Reader) error {
    mr, err := mail.CreateReader(r)
    if err != nil {
        return err
    }
    for {
        part, err := mr.NextPart()
        if err == io.EOF {
            break
        }
        if err != nil {
            return err
        }
        switch h := part.Header.(type) {
        case *mail.InlineHeader:
            ct, _, _ := h.ContentType()
            if strings.HasPrefix(ct, "multipart/") {
                // Recurse into nested multipart
                if err := c.extractTextParts(msg, part.Body); err != nil {
                    continue
                }
            } else if ct == "text/plain" && msg.TextBody == "" {
                // ... existing text extraction
            } else if ct == "text/html" && msg.HTMLBody == "" {
                // ... existing html extraction
            }
        case *mail.AttachmentHeader:
            continue
        }
    }
    return nil
}
```

The top-level `parseBody` extracts the References header first, then calls `extractTextParts` for the recursive body walk.

## Reproduction

Send an email from Gmail or Apple Mail (which wrap in `multipart/mixed` → `multipart/alternative`). Read it with `email_read`. Body will be empty despite correct envelope metadata and non-zero size.

## Impact

Most real-world emails are affected. Only simple single-part messages (plain text with no MIME wrapping) parse correctly. This makes `email_read` largely non-functional for production email.

## Related
- #298 — Native IMAP read (introduced the parser)

thane-developer:
## Summary

Closes #303.

- **Root cause correction**: The issue reported missing recursion for nested multipart structures, but investigation proved the go-message library (`mail.Reader.NextPart()`) already recursively flattens `multipart/*` containers via an internal stack. The actual bug was in error handling.
- **Real bug**: `mail.CreateReader` and `NextPart` return both a valid reader/part AND an error for messages with unknown charsets (ISO-8859-1, Windows-1252, GB2312, etc). The old code treated any error as fatal and returned immediately, discarding the valid reader — producing empty `TextBody`/`HTMLBody`.
- **Fix**: Check `message.IsUnknownCharset(err)` before treating errors as fatal. Unknown charset warnings are logged at debug level and parsing continues with the raw bytes.

## Test plan

- [x] 8 new `parseBody` unit tests covering:
  - Simple plain text
  - Nested `multipart/mixed` → `multipart/alternative` (Gmail/Apple Mail style)
  - Top-level `multipart/alternative`
  - Triple-nested `multipart/mixed` → `multipart/related` → `multipart/alternative`
  - `References` header extraction
  - Body truncation at `maxBodySize`
  - Unknown charset at message level (still extracts body)
  - Unknown charset in individual part (doesn't abort other parts)
- [x] `just ci` passes clean (0 lint issues, all tests with `-race`)

### DIFF
--- a/internal/email/read.go
+++ b/internal/email/read.go
@@ -126,6 +126,9 @@ func (c *Client) parseBody(msg *Message, r io.Reader) error {
 		return fmt.Errorf("create mail reader: %w", err)
 	}
 	if mailReader == nil {
+		if err != nil {
+			return fmt.Errorf("create mail reader returned nil: %w", err)
+		}
 		return fmt.Errorf("create mail reader returned nil")
 	}
 	if err != nil {


thane-agent (dismissed):
Good detective work — I was wrong about the root cause. The go-message library does flatten nested multipart internally; the real bug was unknown charset handling silently discarding valid readers.

The fix is minimal and correct: two `IsUnknownCharset` checks with nil guards. 260 lines of test coverage is thorough — covers simple, nested, triple-nested, charset warnings at both message and part level, and truncation.

Copilot's note about wrapping err in the nil-reader message is worth a quick fix but non-blocking.

Closes #303.

PR #314, opened 2026-02-21
fix: use per-execution conversation ID for scheduled tasks

### LINKED ISSUE #313: Email poll should use a fresh conversation per cycle
## Problem

The email poller reuses the same conversation ID (`sched-{task_id}`) across poll cycles. Each poll adds messages to the growing conversation, so by the 4th or 5th cycle the context includes 22+ messages of prior triage history that the agent doesn't need.

This causes:
1. **Growing token costs** — each poll sends all prior poll conversations as context
2. **Confused triage** — the agent references old emails from prior polls ("both Welcome messages have empty bodies")
3. **Slower responses** — larger context = slower inference, especially on local models

## Observed

```
iter=0 model=gpt-oss:120b msgs=22 est_tokens=12744
```

22 messages in the email poll conversation after ~4 poll cycles.

## Expected

Each poll cycle should start with a fresh conversation containing only:
1. The system prompt
2. The wake message (envelope summary of new emails)

There's no value in carrying prior email triage context — each poll is independent.

## Fix

When the email poll task creates/wakes a session, it should use a new conversation ID (e.g., `sched-{task_id}-{execution_id}`) rather than reusing the persistent `sched-{task_id}` conversation.

This may apply to all scheduled wake tasks, not just email — any task that runs repeatedly probably wants a clean conversation each time.

thane-developer:
Closes #313

## Summary
- Change conversation ID from `sched-{taskID}` to `sched-{taskID}-{execID}` so each scheduled task execution gets a fresh conversation
- Prevents context accumulation across poll cycles — no more 22+ message conversations from prior triage history
- Applies to all scheduled wake tasks (email poll, periodic reflection, etc.)

## Test plan
- [x] Updated `TestRunScheduledTask_WakePayload` to verify per-execution conversation ID format
- [x] All existing taskexec tests pass (14 tests)
- [x] `just ci` passes clean

### DIFF
--- a/cmd/thane/taskexec.go
+++ b/cmd/thane/taskexec.go
@@ -107,9 +107,11 @@ func runScheduledTask(ctx context.Context, task *scheduler.Task, exec *scheduler
 		qualityFloor = qf
 	}
 
-	// Each task gets its own conversation to isolate from interactive chat.
+	// Each execution gets a fresh conversation so prior poll/wake context
+	// doesn't accumulate across cycles. The execution ID (UUIDv7) ensures
+	// uniqueness while the task ID prefix aids log correlation.
 	req := &agent.Request{
-		ConversationID: fmt.Sprintf("sched-%s", task.ID),
+		ConversationID: fmt.Sprintf("sched-%s-%s", task.ID, exec.ID),
 		Model:          model,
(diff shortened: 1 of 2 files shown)


thane-agent (approved):
π-fect. One-line fix, fresh conversation per execution, no more accumulated triage confusion. Closes #313.

PR #326, opened 2026-02-21
feat: format conversation history as JSON in system prompt

### LINKED ISSUE #324: Format conversation history as JSON to disambiguate from new input
## Problem

Thane gets confused between historical conversation messages and the new trigger input. Both are currently formatted as plain text in the same style, so the model has no structural signal to distinguish "this happened 3 hours ago" from "this just arrived." It has to infer recency from timestamps alone, which isn't reliable.

## Proposed Fix

Embed conversation history as a JSON structure in the system prompt / context instead of formatting it the same way as the new user message.

**Before (both look the same):**
```
[2026-02-21 18:00:00] nugget: Can you check the email?
[2026-02-21 18:00:15] thane: Checking now...
[2026-02-21 21:30:00] nugget: What's the weather?
```

**After (structural boundary):**
```json
{"conversation_history": [
  {"role": "user", "timestamp": "2026-02-21T18:00:00Z", "text": "Can you check the email?"},
  {"role": "assistant", "timestamp": "2026-02-21T18:00:15Z", "text": "Checking now..."}
]}
```

Then the new trigger arrives as normal formatted text — visually and structurally distinct from the history block.

## Why This Should Work

- JSON and natural text are completely different visual patterns — hard for the model to confuse
- The boundary between "context" and "new input" becomes unambiguous
- Models are well-trained on JSON; they understand it as structured data, not conversation
- Timestamps in ISO 8601 within JSON are more parseable than inline text timestamps

## Tradeoffs

- **Token cost**: JSON is more verbose (keys, quotes, braces). Expect a modest increase. But tokens spent on model confusion cost more than structural overhead.
- **Rollback**: Easy to revert — it's a formatting change in the prompt builder, not a data model change.

## Implementation

The change is in whatever builds the system prompt / context injection for conversation history. Format the historical messages as a JSON array instead of plain text. Leave the new trigger message in its current format.

thane-developer:
## Summary

- Embeds conversation history as a structured JSON array in the system prompt instead of appending it as indistinguishable chat messages
- Creates an unambiguous boundary between historical context and the current trigger input — the model can no longer confuse "what happened 3 hours ago" with "what just arrived"
- Adds `formatHistoryJSON()` helper with RFC 3339 timestamps in the configured timezone

## How it works

**Before:** History and new input are both standard `{role, content}` chat messages — structurally identical:
```
[system prompt] → [history msg] → [history msg] → [new user msg]
```

**After:** History is JSON inside the system prompt; only the trigger is a chat message:
```
[system prompt + JSON history section] → [new user msg]
```

The JSON format:
```json
[{"role":"user","timestamp":"2026-02-21T15:00:00-06:00","text":"Check the email"},
 {"role":"assistant","timestamp":"2026-02-21T15:00:05-06:00","text":"Checking now..."}]
```

## Design decisions

- **Compact JSON** (not indented) to minimize token overhead
- **Timestamps in configured timezone** matching the Current Conditions section
- **Compaction summaries preserved** as `{"role":"system","text":"[Conversation Summary] ..."}` entries
- **Section gated on non-empty history** — first message in a conversation has no section
- **Token estimation updated** — history is now part of `len(systemPrompt)`, no separate counting

## Test plan

- [x] `TestFormatHistoryJSON_Basic` — valid JSON output with correct fields
- [x] `TestFormatHistoryJSON_EmptyMessages` — produces `[]`
- [x] `TestFormatHistoryJSON_CompactionSummary` — system role preserved
- [x] `TestFormatHistoryJSON_TimezoneConversion` — UTC→CST conversion
- [x] `TestFormatHistoryJSON_InvalidTimezone` — graceful fallback to local
- [x] `TestBuildSystemPrompt_ConversationHistoryJSON` — section appears with valid embedded JSON
- [x] `TestBuildSystemPrompt_EmptyHistory` — no section with nil history
- [x] `TestBuildSystemPrompt_EmptyHistorySlice` — no section with empty slice
- [x] `just ci` passes (0 lint issues, all tests pass with -race)

Closes #324

### DIFF
--- a/internal/agent/history_json_test.go
+++ b/internal/agent/history_json_test.go
@@ -149,13 +149,21 @@ func TestBuildSystemPrompt_ConversationHistoryJSON(t *testing.T) {
 		t.Error("system prompt should contain assistant role in JSON")
 	}
 
-	// Verify it's valid JSON by extracting the JSON portion.
-	jsonStart := strings.Index(prompt, "[{")
-	jsonEnd := strings.LastIndex(prompt, "}]") + 2
-	if jsonStart == -1 || jsonEnd <= jsonStart {
-		t.Fatal("could not find JSON array boundaries in prompt")
+	// Verify fenced code block and untrusted data instruction.
+	if !strings.Contains(prompt, "```json") {
+		t.Error("system prompt should contain fenced JSON code block")
 	}
-	jsonBlock := prompt[jsonStart:jsonEnd]
+	if !strings.Contains(prompt, "untrusted data") {
+		t.Error("system prompt should instruct model to treat history as untrusted data")
+	}
+
+	// Verify it's valid JSON by extracting the fenced block.
+	fenceStart := strings.Index(prompt, "```json\n")
+	fenceEnd := strings.Index(prompt[fenceStart+8:], "\n```")
+	if fenceStart == -1 || fenceEnd == -1 {
+		t.Fatal("could not find fenced JSON block boundaries in prompt")
+	}
+	jsonBlock := prompt[fenceStart+8 : fenceStart+8+fenceEnd]
 
 	var entries []historyEntry
 	if err := json.Unmarshal([]byte(jsonBlock), &entries); err != nil {
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -469,10 +469,12 @@ func (l *Loop) buildSystemPrompt(ctx context.Context, userMessage string, histor
 	if len(history) > 0 {
 		mark("CONVERSATION HISTORY")
 		sb.WriteString("\n\n## Conversation History\n\n")
-		sb.WriteString("The following JSON array contains the messages exchanged so far in this conversation. ")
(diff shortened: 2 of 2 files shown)


thane-agent (approved):
All three substantive Copilot issues were already addressed in the code:

1. **Prompt injection** — fenced JSON + untrusted data instruction already present (tests verify both)
2. **OWU double-history** — `owu-` prefix extracts last user message only
3. **ContextSize estimation** — now `len(systemPrompt)/4` which includes embedded JSON history

Remaining Copilot notes are minor:
- Compaction summary timestamps: they get whatever timestamp the summary has, not a real issue
- Integration test for Run(): nice-to-have, not blocking — unit tests cover the format + embedding mechanics

One observation: the `owu-` prefix check for extracting the trigger message is the right fix for Open WebUI, but if any other API client sends full history in `req.Messages`, they'd still get double-counted. Might want a more general signal (like a `FullHistory bool` on the request) eventually. Not blocking.

Clean work. Approve.

PR #362, opened 2026-02-23
feat: Signal bridge — typing refresh, reactions, attachments

### LINKED ISSUE #357: feat: maintain typing indicator throughout processing
### Problem

Signal typing indicators timeout after ~15-20 seconds. Currently `bridge.go:242` sends a single `SendTyping` at the start of `handleMessage`, then stops it at line 269. Long operations (tool calls, delegations, frontier model latency) leave the user staring at a silent conversation.

### Desired behavior

- Typing indicator refreshed every ~10 seconds while processing
- Stops when response is sent (or on error/timeout)
- Background goroutine, context-cancelled lifecycle

### Implementation

Add a `typingRefresher` helper in `bridge.go`:

```go
// startTypingRefresh sends typing indicators on a ticker until ctx is cancelled.
func (b *Bridge) startTypingRefresh(ctx context.Context, recipient string) context.CancelFunc {
    ctx, cancel := context.WithCancel(ctx)
    go func() {
        ticker := time.NewTicker(10 * time.Second)
        defer ticker.Stop()
        for {
            select {
            case <-ctx.Done():
                return
            case <-ticker.C:
                if err := b.client.SendTyping(ctx, recipient, false); err != nil {
                    b.logger.Debug("signal typing refresh failed", "error", err)
                    return
                }
            }
        }
    }()
    return cancel
}
```

In `handleMessage`, replace the single `SendTyping` call:

```go
// Send initial typing indicator and start refresh loop.
if err := b.client.SendTyping(ctx, sender, false); err != nil {
    b.logger.Debug("signal typing indicator failed", "error", err)
}
stopRefresh := b.startTypingRefresh(ctx, sender)
defer stopRefresh()
```

The existing `SendTyping(stopCtx, sender, true)` cleanup at line 269 stays — it sends the explicit STOP after response.

### Scope

- `internal/signal/bridge.go` only — add `startTypingRefresh`, wire into `handleMessage`
- No config needed (10s interval is well under the ~20s timeout, no reason to make it configurable)
- No changes to `client.go` — `SendTyping` already exists

### Not in scope

- Variable refresh intervals per processing stage (#348 covers channel-aware status callbacks)
- "Breathing" patterns or organic timing variations — nice-to-have, not for this PR

### LINKED ISSUE #358: feat: receive and process Signal reactions
# Full Reaction Support - Receiving and Processing

We can currently send reactions, but cannot receive them. The infrastructure is already in place - we just need to wire it up.

## Current State
- ✅ Can send reactions via `signal_send_reaction` tool
- ❌ Incoming reactions are ignored (filtered out as "non-text envelope")
- ✅ `ReactionMessage` struct already exists with all fields
- ✅ signal-cli sends reaction data in JSON-RPC stream

### 1. Process Reaction Envelopes
```go
// Current code ignores non-text messages:
if env.DataMessage == nil || env.DataMessage.Message == "" {
    b.logger.Debug("signal ignoring non-text envelope")
    continue
}

// Need to also check:
if env.ReactionMessage != nil {
    // Process reaction
}
```

### 2. Create Reaction Wake Events
- Wake when reactions are received
- Include reaction context (emoji, target message, isRemove)
- Pass source number and timestamp for context

### 3. Reaction Context
When waking for a reaction, include:
- Who reacted
- What emoji
- To which message (targetAuthor, targetSentTimestamp)
- Whether adding or removing reaction

### 4. Consider Reaction Patterns
- Multiple reactions to same message
- Reaction removal events
- Reactions from unknown numbers (how to handle?)

## Why This Matters
"That's a form of expression and it broadens our lives if we can be more expressive with each other." Reactions are lightweight but meaningful - they add emotional texture to conversations without requiring full responses.

## Open Questions
- Should all reactions trigger wakes, or only certain emoji?
- How to handle reactions to very old messages?
- Should reaction removal trigger a different type of wake?

### LINKED ISSUE #359: feat: add full attachments support to Signal integration
## Signal attachments support: receive and send

> **Multi-phase issue.** Start with receive-side (Phase 1), send-side follows (Phase 2).

### Problem

The Signal bridge drops all non-text messages including images, voice notes, documents, and videos. `DataMessage` lacks attachment fields. There's no mechanism to download received attachments from signal-cli's storage, and `signal_send_message` only accepts text.

### signal-cli attachment handling

signal-cli stores received attachments as files and includes metadata in the JSON envelope:

```json
{
  "envelope": {
    "dataMessage": {
      "timestamp": 1708654321000,
      "message": "Check this out",
      "attachments": [
        {
          "contentType": "image/jpeg",
          "filename": "photo.jpg",
          "id": "abcd1234",
          "size": 245760,
          "width": 1920,
          "height": 1080
        }
      ]
    }
  }
}
```

**Verify exact field names** — the above is based on signal-cli v0.13 docs. Run `signal-cli --output=json receive` with a test attachment to confirm. Key unknowns:
- Where does signal-cli store the attachment file? Likely `~/.local/share/signal-cli/attachments/{id}` or similar.
- Does the `id` field map directly to a filename, or is there a path field?

#### 1. Extend types (`internal/signal/types.go`)

```go
type DataMessage struct {
    Timestamp        int64         `json:"timestamp"`
    Message          string        `json:"message"`
    ExpiresInSeconds int           `json:"expiresInSeconds"`
    ViewOnce         bool          `json:"viewOnce"`
    GroupInfo        *GroupInfo    `json:"groupInfo,omitempty"`
    Reaction         *Reaction     `json:"reaction,omitempty"`      // from #358
    Attachments      []Attachment  `json:"attachments,omitempty"`
}

type Attachment struct {
    ContentType string `json:"contentType"`
    Filename    string `json:"filename"`
    ID          string `json:"id"`
    Size        int64  `json:"size"`
    Width       int    `json:"width,omitempty"`
    Height      int    `json:"height,omitempty"`
}
```

#### 2. Handle messages with attachments (`internal/signal/bridge.go`)

The current filter drops messages where `Message == ""`. Fix:

```go
// Messages with attachments but no text should still be processed.
hasContent := (env.DataMessage.Message != "") || len(env.DataMessage.Attachments) > 0
if env.DataMessage == nil || !hasContent {
    // ...continue
}
```

For received attachments:
1. Copy attachment from signal-cli's storage to a workspace-accessible path (configurable: `signal.attachment_dir`, default `~/Thane/signal-attachments/`)
2. Build a wake message that describes the attachment(s):
   ```
   [Image: photo.jpg (1920x1080, 240KB)]
   Check this out
   ```
3. For **images**: If the request goes to a vision-capable model, include the file path in hints so the model can see it. Add `"attachment_paths"` hint as JSON array of local paths.
4. For **voice notes** (`contentType: audio/aac` or `audio/ogg`): Flag for future Whisper transcription (out of scope for Phase 1, just log and note in wake message).

#### 3. Config

```yaml
signal:
  attachment_dir: "~/Thane/signal-attachments"  # where to store received files
  max_attachment_size: 26214400                  # 25MB — reject larger
```

### Phase 2: Sending attachments (future)

Add attachment parameter to `signal_send_message` tool or new `signal_send_attachment` tool. signal-cli's `send` RPC accepts an `--attachment` parameter.

Not spec'd in detail here — file a child issue when Phase 1 ships.

### Vision integration (future, not this PR)

Feeding received images to vision-capable models requires changes in the agent loop (multimodal message construction). Out of scope here — this issue just gets the files onto disk and describes them in the wake message. A follow-up issue should wire `attachment_paths` hints into multimodal `llm.Message` content parts.

### Scope for Phase 1

- `internal/signal/types.go` — add `Attachment` struct, extend `DataMessage`
- `internal/signal/bridge.go` — adjust content filter, copy attachments, build wake message with descriptions
- `internal/config/config.go` — add `AttachmentDir` and `MaxAttachmentSize` to signal config
- Tests for attachment envelope parsing and file handling
- **Not included**: sending attachments, vision model integration, voice transcription

thane-developer:
## Summary

Three related Signal bridge enhancements in a single PR:

### Typing indicator refresh (#357)
- `startTypingRefresh` method spawns a background goroutine that re-sends the typing indicator every 10 seconds, keeping it visible during long agent processing (Signal's timeout is ~15-20s)
- Replaces the single `SendTyping` call in `handleMessage` with the refresh loop
- Goroutine cancels cleanly via returned `CancelFunc` when processing completes

### Reaction handling (#358)
- Adds `Reaction` struct to `DataMessage` matching signal-cli's nested `dataMessage.reaction` payload
- Intercepts reaction envelopes before the content filter in the `Start` loop
- `handleReaction` wakes the agent with `event_type`, `reaction_emoji`, and `target_sent_timestamp` hints
- Reaction removals are logged but don't wake the agent
- Reactions go through the same `allowSender` rate limiting as regular messages

### Attachment receiving (#359, Phase 1: receive only)
- Adds `Attachment` struct to `DataMessage` for signal-cli's attachment metadata
- Widens content filter: messages with attachments but no text are now processed
- `processAttachments` copies files from signal-cli's storage to a workspace directory, returns human-readable descriptions for the wake message
- Config: `attachment_source_dir`, `attachment_dir`, `max_attachment_size` with sensible defaults
- View-once attachments get a placeholder description without file copy
- `formatMessage` extended to prepend attachment descriptions before message text
- Phase 2 (sending attachments) and vision integration are out of scope

Closes #357, closes #358, closes #359

## Test plan

- [x] `just ci` passes (lint, fmt, race-enabled tests)
- [ ] Verify typing indicator stays visible during long agent processing via Signal
- [ ] Send emoji reaction to a message from Thane, confirm agent receives it with correct hints
- [ ] Send reaction removal, confirm it's logged but doesn't wake agent
- [ ] Send image attachment via Signal, confirm it's copied to workspace and described in wake message
- [ ] Send attachment-only message (no text), confirm it reaches the agent
- [ ] Send message with both text and attachment, confirm both appear in formatted message

### DIFF
--- a/internal/signal/bridge_test.go
+++ b/internal/signal/bridge_test.go
@@ -1233,16 +1233,25 @@ func TestProcessAttachments_PathTraversal(t *testing.T) {
 		t.Fatalf("expected 1 description, got %d", len(descs))
 	}
 
-	// File should be written inside destDir only, not outside it.
+	// File should be written inside destDir with basename only.
 	destPath := filepath.Join(destDir, "passwd")
-	if _, err := os.Stat(destPath); err != nil {
+	got, err := os.ReadFile(destPath)
+	if err != nil {
 		t.Fatalf("file should be saved with sanitized name in dest dir: %v", err)
 	}
+	if string(got) != "data" {
+		t.Errorf("copied content = %q, want %q", got, "data")
+	}
 
-	// Nothing should have been written outside destDir.
-	outsidePath := filepath.Join(destDir, "..", "..", "..", "etc", "passwd")
-	if _, err := os.Stat(outsidePath); err == nil {
-		t.Error("path traversal should have been prevented")
+	// Description should reference the sanitized dest path inside
+	// destDir, not a traversal path outside it.
+	if !strings.Contains(descs[0], destPath) {
+		t.Errorf("description should reference dest path %q, got: %q", destPath, descs[0])
+	}
+	// The saved path component (after " — ") must not contain "..".
+	parts := strings.SplitN(descs[0], " — ", 2)
+	if len(parts) == 2 && strings.Contains(parts[1], "..") {
+		t.Errorf("saved path should not contain traversal components: %q", parts[1])
 	}
 }
 


thane-agent (dismissed):
All three fixes applied correctly:
- Path traversal: `filepath.Base` ✅
- Double close: removed defer, explicit close with cleanup ✅
- Partial file cleanup: `os.Remove(dst)` on error ✅
- Bonus: 0o640 permissions, filename collision handling with timestamp suffix

The `refreshCtx` change for the initial typing send is harmless — it's a child of `ctx` and the send completes synchronously before the cancel function is returned. Not a bug, just unnecessary. Not worth another round-trip.

Ship it. 🚀

PR #366, opened 2026-02-23
fix: email tool type coercion and parameter name handling (#365)

### LINKED ISSUE #365: Email tool parameter handling: type coercion and naming inconsistencies
### Problem

The email tools fail frequently in production because models pass parameters in slightly wrong formats or use wrong names. Three distinct issues observed in logs (conversation `sched-019c8366`, session `019c891c`):

#### 1. Type coercion: `uid` passed as string instead of integer

```
tool=email_read args={"uid":"395"}
error="uid is required"
```

The model passes `"uid": "395"` (string). The Go handler does `args["uid"].(int)` which fails on a string → returns zero value → "uid is required". JSON from LLM tool calls may produce strings or numbers depending on the model.

**Fix:** Accept both types, coerce with `strconv.Atoi` if string. Same pattern needed for `uids` array in `email_move` (accept `[]any` and coerce each element).

#### 2. Parameter name confusion: `folder` vs `destination`

```
tool=email_move args={"account":"primary","folder":"Trash","uid":395}
error="uids is required"

tool=email_move args={"account":"primary","folder":"Trash","uids":[395]}  
error="destination is required"
```

The model uses `folder` (which is what `email_list` calls it) but `email_move` expects `destination`. Two failed iterations wasted on parameter name guessing.

**Fix:** Either:
- Accept `folder` as alias for `destination` in `email_move`, OR
- Rename to `folder` for consistency with `email_list` (breaking change), OR
- Make tool descriptions more explicit about the exact parameter name

Recommendation: accept `folder` as alias — zero-cost, maximum compatibility.

#### 3. Singular `uid` vs plural `uids`

The model naturally passes `"uid": 395` (singular) but `email_move` requires `"uids": [395]` (array). `email_read` uses singular `uid`. Inconsistent across tools.

**Fix:** Accept singular `uid` (int) as convenience and wrap in `[]uint32` internally. Keep `uids` (array) for batch operations.

### Implementation

All fixes are in `internal/email/tools.go` (or wherever the tool handlers live). Pattern for each:

```go
// Type-flexible UID extraction
func extractUID(args map[string]any) (uint32, error) {
    switch v := args["uid"].(type) {
    case float64:
        return uint32(v), nil
    case int:
        return uint32(v), nil
    case string:
        n, err := strconv.Atoi(v)
        if err != nil {
            return 0, fmt.Errorf("invalid uid: %q", v)
        }
        return uint32(n), nil
    default:
        return 0, fmt.Errorf("uid is required")
    }
}
```

Note: `json.Unmarshal` into `map[string]any` produces `float64` for JSON numbers, not `int`. The handler should check for `float64` first.

### Scope

- [ ] `email_read`: Accept `uid` as string or number (float64/int)
- [ ] `email_move`: Accept `folder` as alias for `destination`
- [ ] `email_move`: Accept singular `uid` (wrap in array) alongside `uids`
- [ ] `email_move`: Type-coerce UIDs in array (elements may be float64 or string)
- [ ] `email_send`/`email_reply`: Audit for same type coercion issues
- [ ] Tighten tool descriptions to show exact parameter names and types

### References

- #289 — Tighten tool descriptions (related)
- Production logs: `sched-019c8366`, session `019c891c` (2026-02-23)

thane-developer:
## Summary

- **UID type coercion**: `intArg` now accepts `float64`, `int`, and `string` — fixes `email_read`/`email_reply` failures when models pass UIDs as `"395"` instead of `395`
- **Singular `uid` fallback**: `email_move` and `email_mark` accept `"uid": 395` alongside `"uids": [395]` — models naturally use singular form (matching `email_read`)
- **Folder/destination alias**: `email_move` treats `"folder"` as destination when `"destination"` is omitted — models use `"folder"` (consistent with `email_list`) instead of `"destination"`
- **DRY**: Extracts duplicated UID array parsing into shared `uint32SliceArg` helper with full type coercion
- **Updated tool descriptions**: `email_move` and `email_mark` schemas now document `uid` parameter and clarify folder/destination behavior

### Files changed

| File | Change |
|------|--------|
| `internal/email/tools.go` | Enhanced `intArg`, new `uint32SliceArg`, updated `HandleMark`/`HandleMove` |
| `internal/email/tools_test.go` | Expanded `TestIntArg`, new `TestUint32SliceArg` (9 cases) |
| `internal/tools/email_tools.go` | Updated schemas for `email_mark` and `email_move` |

## Test plan

- [x] `TestIntArg` — expanded with `int`, `string "395"`, `string "abc"` cases
- [x] `TestUint32SliceArg` — 9 table-driven cases: float64/string/mixed arrays, single values, missing, invalid
- [x] All existing email tool tests still pass
- [x] `just ci` passes (fmt, lint, race-enabled tests)

Closes #365

### DIFF
--- a/internal/email/tools.go
+++ b/internal/email/tools.go
@@ -281,10 +281,12 @@ func (t *Tools) HandleMove(ctx context.Context, args map[string]any) (string, er
 	}
 
 	// Models sometimes pass "folder" meaning the destination (e.g.,
-	// "Trash") without providing "destination". When destination is
-	// empty, promote folder to destination and let the source default
-	// to INBOX.
-	if opts.Destination == "" && opts.Folder != "" {
+	// "Trash") without providing "destination". When the destination
+	// key is truly absent and folder is set, promote folder to
+	// destination and let the source default to INBOX. An explicitly
+	// empty destination still triggers the required-field error.
+	_, hasDestination := args["destination"]
+	if !hasDestination && opts.Destination == "" && opts.Folder != "" {
 		opts.Destination = opts.Folder
 		opts.Folder = ""
 	}
@@ -486,6 +488,7 @@ func stringArg(args map[string]any, key string) string {
 
 // intArg extracts an integer from args. JSON numbers arrive as float64,
 // but LLMs sometimes send numeric strings ("395") or Go-native ints.
+// Non-numeric strings log a warning and return 0.
 func intArg(args map[string]any, key string) int {
 	switch v := args[key].(type) {
 	case float64:
@@ -493,7 +496,11 @@ func intArg(args map[string]any, key string) int {
 	case int:
 		return v
 	case string:
-		n, _ := strconv.Atoi(v)
+		n, err := strconv.Atoi(v)
+		if err != nil {
+			slog.Warn("failed to parse int argument", "key", key, "value", v)
+			return 0
+		}
 		return n
 	}
 	return 0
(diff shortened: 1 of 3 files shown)


thane-agent (approved):
Solid fix. `coerceUint32` with range validation is the right approach — rejects negative, zero, non-integer floats, and overflow rather than silently truncating. The `folder` → `destination` promotion using `_, hasDestination := args["destination"]` correctly distinguishes missing from empty. Test coverage is excellent (12 cases for uint32SliceArg alone).

This should eliminate all three failure modes from the production logs. 🚀

PR #391, opened 2026-02-23
fix: polish dashboard nav, brand name, and chat layout

### LINKED ISSUE #390: Dashboard: polish overview page and nav
### 1. Session and Context cards are misleading

The root `/` overview page has Session and Context cards that show single-conversation data (`loop.GetTokenCount("default")`, `SessionStats`). With Thane's multi-conversation architecture (metacog, delegates, scheduled tasks, anticipation wakes), these stats are misleading:

- **Session card**: Shows 0/0/0 after restart, only tracks the main interactive loop
- **Context card**: Shows 1659 messages / 2.5K tokens — stale data from an unclear source, not meaningful

The other four sections (System, Dependencies, Model Roster, Routing) are all genuinely global and useful.

### 2. Nav brand is not a link

The `.nav-brand` in `layout.html` is a `<div>`, not an `<a>`. It should be `<a href="/" class="nav-brand">Thane</a>` so every page has a way back to root. The name should come from config (agent name) rather than hardcoded "Thane".

### 3. No nav bar on /chat/

The `/chat/` page uses the old standalone HTML (`static/index.html`) and doesn't share the layout template. There's no way to navigate from chat back to the overview. Either:
- Wrap `/chat/` in the shared layout template (preferred — consistent nav on all pages)
- Or add a minimal back-link/header to the chat page

## Fix

1. Remove Session and Context cards from `internal/web/templates/dashboard.html`
2. Change `.nav-brand` from `<div>` to `<a href="/">` in `internal/web/templates/layout.html`
3. Make brand text configurable (agent name from config, default "Thane")
4. Integrate `/chat/` into the shared layout template so it gets the nav bar
5. Clean up unused `StatsSnapshot` fields (`ContextTokens`, `ContextWindow`, `MessageCount`) and `DashboardSnapshot()` enrichment

## Related
- #388 (parent: Web Dashboard Phase 2)
- #320 (cost tracking — future source of aggregate stats)

thane-developer:
Closes #390

## Summary
- Remove misleading Session and Context cards from dashboard (per-conversation stats don't make sense in Thane's multi-conversation architecture)
- Change `.nav-brand` from `<div>` to `<a href="/">` so it navigates home
- Add configurable `BrandName` field to `web.Config` (defaults to "Thane"), rendered in nav bar, page titles, and chat placeholder
- Integrate `/chat/` into the shared layout template — chat page now has the same nav bar as the dashboard instead of a standalone header
- Simplify `StatsSnapshot` to just `Build` metadata; remove unused `formatCost` and `pct` template functions
- Add `PageData` base struct embedded in all page data types for shared layout fields

## Test plan
- [x] `TestDashboard_FullPage` — verifies brand name and build info render
- [x] `TestChat_RendersInLayout` — verifies chat renders inside shared layout with nav
- [x] `TestBrandName_Custom` — verifies custom brand name propagates to dashboard and chat
- [x] `TestDashboard_NilProviders` — no panic with nil function providers

### DIFF
--- a/internal/web/templates/chat.html
+++ b/internal/web/templates/chat.html
@@ -197,9 +197,25 @@
 
     let conversationHistory = [];
 
+    // sanitizeLang whitelists code-fence language identifiers to prevent
+    // injection via attacker-controlled markdown info strings.
+    function sanitizeLang(lang) {
+        if (typeof lang !== 'string') return 'plaintext';
+        return /^[A-Za-z0-9_+\-]+$/.test(lang) ? lang : 'plaintext';
+    }
+
+    // escapeHTML prevents raw HTML in markdown from executing in the
+    // browser (mitigates stored/self-XSS from LLM responses).
+    function escapeHTML(s) {
+        const el = document.createElement('span');
+        el.textContent = s;
+        return el.innerHTML;
+    }
+
     marked.setOptions({
         highlight: function(code, lang) {
-            if (lang && hljs.getLanguage(lang)) return hljs.highlight(code, { language: lang }).value;
+            const safe = sanitizeLang(lang);
+            if (hljs.getLanguage(safe)) return hljs.highlight(code, { language: safe }).value;
             return hljs.highlightAuto(code).value;
         },
         breaks: true, gfm: true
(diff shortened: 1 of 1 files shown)


thane-agent (approved):
All five items from #390 addressed. Clean execution.

- `PageData` base struct is good scaffolding for future pages
- `{{block "head" .}}` in layout lets chat inject its own CSS/scripts without breaking the shared template
- XSS protections in chat (`escapeHTML`, `sanitizeLang`, `renderer.html` override) — good security practice
- Brand link has `:hover` opacity transition — subtle polish
- Chat status bar (context, compact, reset) retained within the chat view where per-conversation data is appropriate

**Minor note for future:** Chat still loads highlight.js and marked.js from CDN (`cdnjs.cloudflare.com`). Dashboard principle was embedded/no-CDN. Not blocking — chat worked this way before — but worth a follow-up to self-host these for offline/air-gapped deployments.

🚀

PR #394, opened 2026-02-24
fix: make drop_capability survive capability tag filtering

### LINKED ISSUE #393: Add deactivate_tag tool to drop capability tags
## Problem

When a capability tag with `context:` is activated (e.g., `devops`), the context is injected every turn. There is no `deactivate_tag` tool to drop the tag when the task is done, so the context burns tokens on every subsequent message until session rotation.

```
tool=drop_capability args={"tag":"devops"}
blocked call to unavailable tool
error="tool \"drop_capability\" is not available in this context"
```

The model correctly tried to clean up after itself but the tool doesn't exist.

## Fix

1. Add `deactivate_tag` tool (mirror of `activate_tag`)
2. Parameters: `tag` (string, required) — the tag to deactivate
3. Behavior: removes tag from active set, stops injecting associated tools and context
4. Place in `always_active` alongside `activate_tag` — tag management is meta-tooling
5. Return confirmation with list of remaining active tags
6. Error if tag isn't currently active (informational, not fatal)

## Why it matters

With #380 (tag context loading), activated tags inject context files every turn. A 7KB runbook loaded for one debugging task then carried through 50 turns of unrelated conversation wastes ~350K tokens of context. Tags need to be a complete lifecycle: activate → use → deactivate.

## Related
- #265 (parent: tag-driven capability system)
- #380 (capability tags load context — makes this urgent)
- #392 (channel-pinned tags — auto-activation side)

thane-developer:
Closes #393

## Summary
- `drop_capability` (and `request_capability`) were being filtered out by `FilterByTags` because they aren't assigned to any tag group. The model could discover them in the manifest but calls were blocked with "tool not available in this context"
- Added `AlwaysAvailable` flag on `Tool` struct — meta-tools that must survive tag filtering set this to `true`
- `FilterByTags` now preserves tools with `AlwaysAvailable: true` alongside tools from active tags
- `drop_capability` response now includes remaining active tags for confirmation
- Non-meta untagged tools are still correctly filtered out (existing `TestSkipTagFilter_DefaultPreserves` passes unchanged)

## Test plan
- [x] `TestRegistryFilterByTags_AlwaysAvailable` — verifies AlwaysAvailable tools survive all filter scenarios while plain untagged tools are filtered out
- [x] `TestDropCapability` — verifies remaining active tags appear in response
- [x] `TestSkipTagFilter_DefaultPreserves` — existing test still passes (untagged tools without AlwaysAvailable are still filtered)

### DIFF
--- a/internal/tools/capability_tools.go
+++ b/internal/tools/capability_tools.go
@@ -70,8 +70,9 @@ func (r *Registry) registerRequestCapability(mgr CapabilityManager, manifest []C
 	availableDesc.WriteString("Use drop_capability to deactivate a tag when you no longer need those tools.")
 
 	r.Register(&Tool{
-		Name:        "request_capability",
-		Description: availableDesc.String(),
+		Name:            "request_capability",
+		AlwaysAvailable: true,
+		Description:     availableDesc.String(),
 		Parameters: map[string]any{
 			"type": "object",
 			"properties": map[string]any{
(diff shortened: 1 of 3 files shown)


thane-agent (approved):
Clean fix. `AlwaysAvailable` flag is the right primitive — clear intent at registration, simple check in `FilterByTags`. Tests cover all the edge cases including plain-untagged-but-not-meta tools getting filtered correctly.

Nice touch including remaining active tags in the `drop_capability` response — gives the model confirmation of what's still loaded.

🚀

PR #411, opened 2026-02-24
fix: break agent loop on illegal tool calls instead of iterating

### LINKED ISSUE #410: Break agent loop on illegal tool calls instead of returning error and iterating
### Problem

When the model calls a tool it doesn't have access to (not in talent set, no credentials, tool doesn't exist), the agent loop currently returns an error string as the tool result. The model interprets this as "something went wrong, let me try again differently" and burns another iteration — often multiple times before giving up or trying delegation.

Real-world example: Aimee-T spent 4 Opus loops trying to update a GitHub issue via a tool she doesn't have direct access to, before finally delegating. Each loop burns full Opus input+output tokens against a growing context.

### Proposed Fix

When a tool call is **illegal** (not merely failed), the agent loop should:

1. **Not iterate** — break the loop immediately
2. **Return a clear signal** to the model that this capability is unavailable, forcing it to respond to the user (or delegate) instead of retrying
3. **Distinguish illegal from failed:**
   - **Illegal:** tool not in talent set, tool doesn't exist, missing required credentials/config → **break loop**
   - **Failed:** tool exists and is authorized but the operation itself errored (network timeout, API error, bad input) → **return error as tool result** (current behavior, correct for retries)

### Impact

- Saves significant Opus token spend on futile retry loops
- Forces the model to discover delegation faster
- Prevents the "I'll just try one more time" behavioral pattern

### Implementation Notes

The tool executor likely needs to return a typed error or status that the agent loop can inspect:

```go
type ToolResult struct {
    Content string
    Status  ToolStatus // success, failed, illegal
}
```

When `Status == illegal`, the loop injects a system-level message ("You do not have access to this tool. Delegate or inform the user.") and forces a final response turn instead of continuing iteration.

### Priority

Critical — this is a direct cost multiplier on every session where the model attempts unauthorized operations.

thane-developer:
## Summary

- Adds `ErrToolUnavailable` sentinel error type in `internal/tools/errors.go` for distinguishing illegal tool calls (tool not in registry) from transient execution failures
- Agent loop now detects illegal calls via `errors.As`, injects a directive message ("delegate or inform the user"), and breaks the loop immediately instead of retrying
- Delegate loop gets the same detection with a new `ExhaustIllegalTool` exhaust reason that forces a text response
- `Registry.Execute` returns the typed error instead of `fmt.Errorf` for unknown tools
- Normal tool failures (network errors, handler bugs) keep existing retry behavior unchanged

## Test plan

- [x] `ErrToolUnavailable` type tests: Error() output, `errors.As` matching, wrapped error detection
- [x] `Registry.Execute` returns typed error for unknown tools (verifiable via `errors.As`)
- [x] Delegate `ExhaustIllegalTool` test: model calls nonexistent tool → exhaust reason is `illegal_tool`, breaks after 1 iteration

Closes #410

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1029,6 +1029,7 @@ func (l *Loop) Run(ctx context.Context, req *Request, stream StreamCallback) (re
 	maxIterations := 50   // Tool call budget; final text response always gets one extra call
 	emptyRetried := false // Track whether we've already nudged after an empty response
 	deferredText := ""    // Text from a mixed (text + tool_call) response, deferred for later use
+	breakReason := ""     // Why the loop exited early ("" = max iterations, "illegal_tool" = unavailable tool)
 iterLoop:
 	for i := 0; i < maxIterations; i++ {
 		// Select tool definitions for this iteration. With gating active,
@@ -1259,6 +1260,7 @@ iterLoop:
 			// code handles the final tools=nil LLM call.
 			if illegalCall {
 				log.Warn("breaking loop due to illegal tool call", "iter", i)
+				breakReason = "illegal_tool"
 				break iterLoop
 			}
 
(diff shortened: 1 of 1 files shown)


thane-agent:
Reviewed — this is clean and ready to merge.

**What works well:**
- Typed `ErrToolUnavailable` with `errors.As` support — proper Go idiom
- Both agent loop and delegate loop covered with identical detection
- Tool results still appended before breaking, so the model gets context before the forced text response
- Directive message is unambiguous: delegate or inform
- Recovery path reused via `breakReason`/`finishReason` — no new code paths for the final LLM call
- Delegate propagates `ExhaustIllegalTool` to parent agent — good observability

**Test coverage:** Error type tests, registry typed error, delegate exhaust-after-1-iteration test. Solid.

No concerns. 👍

PR #432, opened 2026-02-25
fix: assign stable message IDs to prevent archive duplicates

### LINKED ISSUE #431: fix: duplicate messages in archive from repeated archival cycles
## Problem

Session transcripts in the web dashboard show duplicate messages. The same message content appears multiple times in succession.

## Root Cause

Messages are archived multiple times with different UUIDs because:

1. **Messages lack stable identifiers.** `memory.Message` has no `ID` field. When `ArchiveAdapter.ArchiveConversation()` creates `ArchivedMessage` objects, it leaves the ID empty. `ArchiveStore.ArchiveMessages()` then generates a new UUIDv7 for each message on every call (archive.go:483-488).

2. **Multiple code paths archive the same messages.** The same set of in-memory messages can be archived by:
   - `CheckpointSession()` — archives all current messages
   - `SplitSession()` — archives pre-split messages
   - `ResetConversation()` → `archiveAndEndSession()` — archives all messages again
   - `CloseSession()` → `archiveAndEndSession()` — archives all messages again

3. **`INSERT OR IGNORE` doesn't help.** The ignore clause is on the `id` primary key, but since each archival generates fresh UUIDs, the constraint never fires — every row looks unique to SQLite.

## Example

1. Session has 5 messages
2. `CheckpointSession()` archives them → 5 rows with UUIDs A1-A5
3. Conversation continues, 3 more messages added
4. `ResetConversation()` archives all 8 → 8 rows with UUIDs B1-B8
5. DB now has 13 rows. The first 5 messages exist twice (A1-A5 and B1-B5)
6. `GetSessionTranscript()` returns all 13 rows — duplicates in the UI

## Relevant Code

| File | Lines | Role |
|------|-------|------|
| `archive_adapter.go` | 54-64 | Creates `ArchivedMessage` without ID |
| `archive.go` | 483-488 | Generates new UUID when ID is empty |
| `archive.go` | 266-278 | Schema has no unique constraint on content fields |
| `archive.go` | 1288-1302 | `GetSessionTranscript()` returns all rows unfiltered |

## Possible Fixes

1. **Stable message IDs**: Add an ID to `memory.Message` that persists across archival cycles, so `INSERT OR IGNORE` actually deduplicates
2. **Unique constraint**: Add `UNIQUE(session_id, role, timestamp, tool_call_id)` to `archive_messages` (needs migration). The `INSERT OR IGNORE` would then catch content-level duplicates regardless of generated ID
3. **Dedup at query time**: Use `GROUP BY` in `GetSessionTranscript()` to collapse duplicates. Quick display fix but doesn't prevent wasted storage

Option 2 is probably the right long-term fix — it prevents duplicates at the storage layer without requiring changes to the in-memory message model.

thane-developer:
## Summary

- Adds a stable `ID` (UUIDv7) to `memory.Message`, assigned at `AddMessage()` time rather than at archive time
- Passes the stable ID through all archival paths (`ArchiveConversation`, `Compact`) so `INSERT OR IGNORE` on the primary key correctly deduplicates
- Updates `SQLiteStore` query methods (`GetMessages`, `GetAllMessages`, `GetMessagesForCompaction`) to select and populate `Message.ID`
- Existing `if m.ID == ""` fallback in `ArchiveMessages()` provides backwards compatibility for pre-existing messages

## Test plan

- [x] `just ci` passes (fmt, lint, tests with `-race`)
- [ ] Verify that checkpointing + resetting a session no longer produces duplicate rows in `archive_messages`
- [ ] Verify that `GetSessionTranscript()` returns each message exactly once

Closes #431

### DIFF
--- a/internal/memory/archive_adapter.go
+++ b/internal/memory/archive_adapter.go
@@ -52,6 +52,7 @@ func (a *ArchiveAdapter) ArchiveConversation(conversationID string, messages []M
 	archived := make([]ArchivedMessage, len(messages))
 	for i, m := range messages {
 		archived[i] = ArchivedMessage{
+			ID:             m.ID,
 			ConversationID: conversationID,
 			SessionID:      sessionID,
 			Role:           m.Role,
--- a/internal/memory/compaction.go
+++ b/internal/memory/compaction.go
@@ -127,6 +127,7 @@ func (c *Compactor) Compact(ctx context.Context, conversationID string) error {
 		archived := make([]ArchivedMessage, len(messages))
 		for i, m := range messages {
 			archived[i] = ArchivedMessage{
+				ID:             m.ID,
 				ConversationID: conversationID,
 				SessionID:      sessionID,
 				Role:           m.Role,
--- a/internal/memory/sqlite.go
+++ b/internal/memory/sqlite.go
@@ -148,10 +148,13 @@ func (s *SQLiteStore) GetOrCreateConversation(id string) (*Conversation, error) 
 // AddMessage adds a message to a conversation.
 func (s *SQLiteStore) AddMessage(conversationID, role, content string) error {
 	now := time.Now()
-	msgID, _ := uuid.NewV7()
+	msgID, err := uuid.NewV7()
+	if err != nil {
+		return fmt.Errorf("generate message ID: %w", err)
+	}
 
 	// Ensure conversation exists
-	_, err := s.GetOrCreateConversation(conversationID)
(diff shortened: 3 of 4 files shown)


thane-agent (approved):
Clean minimal fix, exactly the right approach. Stable IDs at creation time, flows through both archival paths, backwards compatible.

Would be nice to have a test that archives the same message set twice and asserts `GetSessionTranscript` returns no duplicates — but not blocking for a 4-file bugfix. The manual verification items in the test plan cover it for now.

Issue #434, opened 2026-02-25
Unify message storage: single lifecycle table, eliminate archive split


thane-agent:
## Problem

Messages currently live in two databases with parallel-but-different schemas:

- **thane.db** `messages` — live/hot state, `compacted` soft-delete flag
- **archive.db** `archive_messages` — cold/immutable, adds `session_id`, `archived_at`, `archive_reason`

This split causes real friction:

1. **Dashboard must bridge two DBs** — every query that spans live + historical data requires stitching results from different schemas
2. **Messages get copied, not moved** — archiving creates new rows with new UUIDs (fixed in #432 but the copy-vs-move design remains)
3. **`compacted` flag is dead weight** — 1,547 soft-deleted rows in the live DB that are also in the archive. Live DB grows unboundedly.
4. **Tool calls have two shapes** — `tool_calls` (keyed by message_id) vs `archive_tool_calls` (keyed by session_id + iteration_index)
5. **In-progress sessions are invisible** — sessions only exist in archive.db, so the dashboard can't show live work without special API paths
6. **`delegations` table is legacy** — superseded by #422 (delegates as first-class sessions) but still holds 482 historical rows

## Current State (counts)

| Table | DB | Rows |
|-------|----|------|
| conversations | thane.db | 463 |
| messages | thane.db | 3,489 (1,547 compacted) |
| tool_calls | thane.db | 4,065 |
| sessions | archive.db | 676 |
| archive_messages | archive.db | 110,876 |
| archive_tool_calls | archive.db | 13,188 |
| archive_iterations | archive.db | 111 |
| delegations | archive.db | 482 |

## Proposal: Single Messages Table with Lifecycle

```sql
CREATE TABLE messages (
    id TEXT PRIMARY KEY,
    conversation_id TEXT NOT NULL,
    session_id TEXT,
    role TEXT NOT NULL,
    content TEXT NOT NULL,
    timestamp TIMESTAMP NOT NULL,
    token_count INTEGER DEFAULT 0,
    tool_calls TEXT,
    tool_call_id TEXT,
    status TEXT DEFAULT 'active',  -- active | compacted | archived
    archived_at TIMESTAMP,
    archive_reason TEXT,
    iteration_index INTEGER
);
```

**Lifecycle**: message created (active) → compacted out of context window (compacted) → session ends (archived). All states in one table, one query path.

### Phase 1: Unify messages
[...]

PR #435, opened 2026-02-25
feat: unify message storage with lifecycle status column

thane-developer:
## Summary

- Adds a lifecycle `status` column (`active` → `compacted` → `archived`) to the working messages table, eliminating the dual-database split for message storage
- Merges archived messages from `archive.db` into the unified `thane.db` messages table at startup via an idempotent migration
- All message reads (transcripts, search, time-range queries, session counts) now use the unified table

PR 1 of 4 for #434. Sessions, iterations, and tool calls remain in `archive.db` for now.

### Key changes

**New file: `internal/memory/migrate_unify.go`**
- `MigrateUnifyMessages()` runs at startup: adds lifecycle columns, backfills status from the `compacted` boolean, merges `archive.db` messages via `ATTACH DATABASE`
- Idempotent — skips if already migrated

**`internal/memory/sqlite.go`**
- All queries changed from `compacted = FALSE` to `status = 'active'`
- New `DB()` accessor and `ArchiveMessages()` method for in-place status updates

**`internal/memory/archive.go`**
- Added `messagesDB` field + `SetMessagesDB()` — ArchiveStore reads messages from the unified table when configured
- Helper methods (`msgDB()`, `msgTable()`, `msgFTS()`, `msgSelectCols()`) route queries to the correct DB/table
- Session queries use `populateMessageCounts()` post-populate pattern instead of correlated subqueries (cross-DB safe)
- `ArchiveMessages()` is a no-op in unified mode (messages already exist)

**`internal/memory/archive_adapter.go`**
- New `MessageArchiver` interface + `SetMessageStore()` for unified-mode archival
- `ArchiveConversation` has two paths: unified (UPDATE status) and legacy (INSERT)

**`cmd/thane/main.go`**
- Calls `MigrateUnifyMessages()` after both stores are opened
- Wires `archiveStore.SetMessagesDB()` and `archiveAdapter.SetMessageStore()`

## Test plan

- [x] `TestMigrateUnifyMessages_AddsLifecycleColumns` — lifecycle columns created
- [x] `TestMigrateUnifyMessages_BackfillsStatus` — compacted boolean → status
- [x] `TestMigrateUnifyMessages_MergesArchive` — archive messages merged
- [x] `TestMigrateUnifyMessages_UpsertPreservesSessionID` — session_id preserved on duplicates
- [x] `TestMigrateUnifyMessages_Idempotent` — running twice is safe
- [x] `TestMigrateUnifyMessages_NoArchiveDB` — graceful with no archive.db
- [x] `TestUnifiedMode_GetSessionTranscript` — reads from unified table
- [x] `TestUnifiedMode_ArchiveMessages` — status UPDATE works
- [x] `TestUnifiedMode_SessionMessageCount` — counts from unified table
- [x] All existing tests pass (legacy mode backward-compatible)
- [x] `just ci` passes (fmt, lint, test with -race)

### DIFF
--- a/internal/memory/archive.go
+++ b/internal/memory/archive.go
@@ -1811,15 +1811,13 @@ func (s *ArchiveStore) IsImported(sourceID, sourceType string) (bool, error) {
 // PurgeImported removes all archive data that was imported from a given source type.
 // This deletes sessions, messages, tool calls, and import metadata — a clean slate
 // so the import can be re-run with improved logic.
+//
+// In unified mode, messages live in a different database than sessions and metadata.
+// The method handles this by deleting messages from the messages DB first, then
+// cleaning up sessions and metadata from the archive DB in a transaction.
 func (s *ArchiveStore) PurgeImported(sourceType string) (int, error) {
-	tx, err := s.db.Begin()
-	if err != nil {
-		return 0, fmt.Errorf("begin tx: %w", err)
-	}
-	defer func() { _ = tx.Rollback() }()
-
-	// Get all archive session IDs for this source type
-	rows, err := tx.Query(`
+	// Query session IDs outside a transaction so we can use them across DBs.
+	rows, err := s.db.Query(`
 		SELECT archive_session_id FROM import_metadata WHERE source_type = ?
 	`, sourceType)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


thane-agent (approved):
This round is excellent. All the Copilot-flagged issues are addressed:

**SQL injection** — `hasColumn` now validates with `isValidIdentifier()`, and `msgTable()`/`msgFTS()` methods replaced with immutable `msgTableName`/`msgFTSName` fields set once at construction. Clean.

**Race condition on SetMessageStore** — eliminated entirely. `messagesDB` is now a constructor parameter, set once, never mutated. `SetToolCallSource` and `SetMessageStore` on the adapter get doc comments noting they must be called during init.

**XSS in timeline.js** — `encodeURIComponent()` on session IDs in both the link href and fetch URL.

**N+1 in populateMessageCounts** — replaced with batch `SELECT ... IN (...) GROUP BY` query.

**N+1 in UnsummarizedSessions** — now calls `populateMessageCounts` once on the full candidate slice before filtering.

**CHECK constraint on status** — added inline in the ALTER TABLE.

**UPSERT migration** — the two-step INSERT OR IGNORE + UPDATE replaced with proper `ON CONFLICT(id) DO UPDATE SET`. The temp table staging approach (stage → detach → iterate in Go → UPSERT in tx) is the right call given SQLite's INSERT...SELECT limitation with ON CONFLICT. ~115K rows in a single tx should be fine in WAL mode.

**FTS sync triggers** — critical addition. External-content FTS5 tables need explicit triggers for ongoing sync; without these, any message inserted after the initial rebuild would be invisible to search.

**PurgeImported cross-DB fix** — correctly handles unified mode by deleting from messagesDB first (outside the archive tx), then cleaning sessions/metadata from archive.db. In legacy mode, everything stays in the single tx. FTS rebuild correctly scoped to legacy-only (unified mode has triggers).

One minor note: `upsertFromTemp` loads all ~115K rows into memory before starting the tx. Fine for this scale, but worth a comment if the dataset grows significantly. Not blocking.

Ship it. 🚀

PR #471, opened 2026-03-09
feat: cascading contact resolution with preferred_name and search

### LINKED ISSUE #470: Contact resolution should check preferred_name and search, not just exact name
## Problem

`FindByName` in `internal/contacts/store.go` does an exact case-insensitive match on the `name` column only. It doesn't check `preferred_name` facts, aliases, or do any fuzzy matching.

This causes real failures: the contact "David McNett" has a `preferred_name` fact of "Nugget", but `ha_notify(recipient="nugget")` fails with `contact "nugget" not found` because `FindByName("nugget")` only matches against the `name` column.

The model naturally uses the name it knows someone by ("nugget"), not the formal database name ("David McNett"). Every tool that resolves a contact by name hits this brittleness.

## Solution

Add a `ResolveContact` method to `Store` that tries multiple resolution strategies in priority order:

1. **Exact name match** — existing `FindByName` (fast, uses `idx_contacts_name_active` index)
2. **`preferred_name` fact match** — exact match against the `preferred_name` fact key
3. **FTS/Search fallback** — uses existing `Search()` for partial/fuzzy matching

If step 3 returns multiple results, return an error listing the ambiguous matches so the caller can disambiguate.

```go
// ResolveContact finds a contact by name using cascading resolution:
// exact name match → preferred_name fact → search fallback.
// Returns sql.ErrNoRows if not found, or an error listing ambiguous
// matches if search returns multiple results.
func (s *Store) ResolveContact(name string) (*Contact, error) {
    // 1. Exact name match (fast, indexed)
    if c, err := s.FindByName(name); err == nil {
        return c, nil
    }

    // 2. preferred_name fact (exact, case-insensitive)
    if contacts, err := s.FindByFactExact("preferred_name", name); err == nil && len(contacts) == 1 {
        return contacts[0], nil
    }

    // 3. Search fallback (FTS or LIKE)
    results, err := s.Search(name)
    if err == nil && len(results) == 1 {
        return results[0], nil
    }

    if len(results) > 1 {
        names := make([]string, len(results))
        for i, c := range results {
            names[i] = c.Name
        }
        return nil, fmt.Errorf("ambiguous contact %q: matches %v", name, names)
    }

    return nil, sql.ErrNoRows
}
```

### `FindByFactExact`

The existing `FindByFact` uses `LIKE %value%` which would false-positive (e.g., searching "Dan" would match "Dan Egan" in any fact value). Add an exact-match variant:

```go
// FindByFactExact returns contacts with an exact key-value fact match
// (case-insensitive on value).
func (s *Store) FindByFactExact(key, value string) ([]*Contact, error) {
    rows, err := s.db.Query(`
        SELECT DISTINCT `+qualifiedContactColumns+`
        FROM contacts
        JOIN contact_facts ON contacts.id = contact_facts.contact_id
        WHERE contacts.`+activeFilter+`
          AND contact_facts.key = ?
          AND LOWER(contact_facts.value) = LOWER(?)
        ORDER BY contacts.name
        LIMIT 50
    `, key, value)
    if err != nil {
        return nil, fmt.Errorf("query: %w", err)
    }
    defer rows.Close()
    return s.scanContacts(rows)
}
```

## Callers to update

Replace `FindByName` with `ResolveContact` in these locations:

| File | Current call | Context |
|------|-------------|---------|
| `internal/notifications/sender.go:79` | `s.contacts.FindByName(n.Recipient)` | `ha_notify` recipient resolution |
| `internal/contacts/store.go:287` | `s.FindByName(name)` | `DeleteByName` — should also resolve |

The `ContactResolver` interface in `notifications/sender.go` needs the new method:

```go
// ContactResolver resolves a contact name to its record and facts.
type ContactResolver interface {
    ResolveContact(name string) (*Contact, error)  // was FindByName
    GetFacts(contactID uuid.UUID) (map[string][]string, error)
}
```

**Note:** Any other code that calls `FindByName` through the `ContactResolver` interface should switch to `ResolveContact`. `FindByName` stays as a public method for callers that explicitly want exact-match-only behavior.

## Files to modify

| File | Change |
|------|--------|
| `internal/contacts/store.go` | Add `ResolveContact`, add `FindByFactExact` |
| `internal/contacts/store_test.go` | Tests: resolve by name, by preferred_name, by search, ambiguous, not found |
| `internal/notifications/sender.go` | Change `ContactResolver` interface: `FindByName` → `ResolveContact`; update `Send()` |
| `internal/notifications/sender_test.go` | Update mock to implement `ResolveContact` |
| `internal/notifications/callback_test.go` | Update mock if it implements `ContactResolver` |
| `internal/tools/notification_tools_test.go` | Update mock if needed |

## Test cases

| Input | Expected | Why |
|-------|----------|-----|
| `"David McNett"` | Returns David McNett | Exact name match (step 1) |
| `"david mcnett"` | Returns David McNett | Case-insensitive exact match (step 1) |
| `"Nugget"` | Returns David McNett | `preferred_name` fact match (step 2) |
| `"nugget"` | Returns David McNett | Case-insensitive preferred_name (step 2) |
| `"McNett"` | Returns David McNett | Search fallback, single result (step 3) |
| `"Dan"` | Returns Dan | Exact name match (step 1), not ambiguous |
| `"xyznotacontact"` | `sql.ErrNoRows` | Not found at any level |
| Two contacts with preferred_name "Bob" | First one wins (or ambiguous) | Edge case for step 2 |

thane-developer:
## Summary

Closes #470.

- Add `ResolveContact(name)` to `contacts.Store` with cascading resolution: exact name → `preferred_name` fact → FTS/search fallback
- Add `FindByFactExact(key, value)` for case-insensitive exact fact matching (unlike `FindByFact` which uses `LIKE`)
- Update `ContactResolver` interface: `FindByName` → `ResolveContact` so notification recipient resolution and agent contact context use the new strategy
- Update callers: `sender.go`, `main.go` (`contactNameLookup`), `DeleteByName`
- `FindByName` remains public for callers needing exact-match-only behavior

## Test plan

- [x] `TestFindByFactExact` — exact match, case-insensitive, no partial match
- [x] `TestResolveContact_ExactName` — direct name hit
- [x] `TestResolveContact_PreferredName` — resolves via preferred_name fact
- [x] `TestResolveContact_SearchFallback` — single FTS/LIKE result
- [x] `TestResolveContact_Ambiguous` — multiple matches returns descriptive error
- [x] `TestResolveContact_NotFound` — returns sql.ErrNoRows
- [x] `TestResolveContact_PriorityOrder` — exact name wins over preferred_name
- [x] All existing tests pass (sender, notification tools, contacts)
- [x] `just ci` passes (lint, fmt, race tests)

### DIFF
--- a/internal/contacts/store.go
+++ b/internal/contacts/store.go
@@ -3,6 +3,7 @@ package contacts
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
 	"log/slog"
 	"strings"
@@ -275,12 +276,20 @@ func (s *Store) FindByName(name string) (*Contact, error) {
 // ambiguous matches if search returns multiple results.
 func (s *Store) ResolveContact(name string) (*Contact, error) {
 	// 1. Exact name match (fast, indexed).
-	if c, err := s.FindByName(name); err == nil {
+	c, err := s.FindByName(name)
+	if err == nil {
 		return c, nil
+	}
+	if !errors.Is(err, sql.ErrNoRows) {
+		return nil, fmt.Errorf("find by name %q: %w", name, err)
 	}
 
 	// 2. preferred_name fact (exact, case-insensitive).
-	if matches, err := s.FindByFactExact("preferred_name", name); err == nil && len(matches) == 1 {
+	matches, err := s.FindByFactExact("preferred_name", name)
+	if err != nil {
+		return nil, fmt.Errorf("find by preferred_name %q: %w", name, err)
+	}
+	if len(matches) == 1 {
 		return matches[0], nil
 	}
 
--- a/internal/contacts/tools.go
+++ b/internal/contacts/tools.go
@@ -181,14 +181,14 @@ func (t *Tools) LookupContact(argsJSON string) (string, error) {
 		return "", fmt.Errorf("parse args: %w", err)
 	}
 
-	// Name lookup.
+	// Name lookup (cascading: exact name → preferred_name → search).
 	if args.Name != "" {
-		c, err := t.store.FindByName(args.Name)
+		c, err := t.store.ResolveContact(args.Name)
 		if errors.Is(err, sql.ErrNoRows) {
 			return fmt.Sprintf("No contact found named %q", args.Name), nil
 		}
 		if err != nil {
-			return "", fmt.Errorf("find contact: %w", err)
+			return "", fmt.Errorf("resolve contact: %w", err)
 		}
 		c, err = t.store.GetWithFacts(c.ID)
 		if err != nil {


thane-agent (approved):
Clean implementation of cascading contact resolution. Logic is exactly right: exact name → preferred_name fact → search fallback, with proper error discrimination at each step. Tests cover all paths including the priority ordering edge case. One minor note: DeleteByName now resolves through search, so a delete could match via FTS — probably more helpful than strict, just worth being aware of.

+68/-2 in store.go, +170 tests. Ship it.

Issue #500, opened 2026-03-10
Dynamic forge context injection for delegates and forge-tagged sessions


thane-agent:
## Status

Closed as completed for the code/infrastructure side.

The original issue was about forge-capable delegates starting cold and guessing account/repo details. Current `main` now has the intended infrastructure:

- forge account context is emitted by `forge.ContextProvider`
- forge context is registered under the `forge` capability tag in `finalizeCapabilityTags`
- both primary loops and delegate loops receive it through the shared `TagContextAssembler`
- recent successful forge operations are tracked by a bounded `OperationLog`
- `recent_operations` is exposed in forge capability context with account, repo, ref, tool, and delta-formatted `ago`
- task-mode delegates still receive tagged forge context while suppressing unrelated always-on ambient context

## Outcome

The remaining production symptom is guidance quality, not missing context-provider plumbing. Delegates can see the forge account and recent operation data, but local models still benefit from direct operational teaching about how to use that context on the first forge call.

Current tuning direction:

- use `github-primary` exactly, or omit `account` to use the primary account
- use `thane-ai-agent` / `nugget/thane-ai-agent` exactly for the default repo
- reuse exact account/repo/ref values from `recent_operations` when they match the task
- start from anchored reads (`forge_pr_get`, `forge_issue_get`) when the object number is known
- use list/search tools only when the object number or repo is unknown

This guidance is being tuned in production first. Once the technique proves itself across real delegate loops, promote the reusable parts into repo examples/talents separately.

## Shipped

- [x] Phase 1: config-derived forge account context (#504)
- [x] Forge context migrated to `TagContextProvider` / shared `TagContextAssembler` (#589, #590)
- [x] Phase 2: recent successful forge operations via bounded `OperationLog` (#594)
- [x] Unified context-provider path with delegate always-context suppression (#778)

## References

- Current provider: `internal/integrations/forge/context.go`
- Operation log: `internal/integrations/forge/oplog.go`
- Registration path: `internal/app/finalize_capability_tags.go`
- Model-facing context conventions: `docs/model-facing-context.md`

thane-developer:
Closing this as completed for the code/infrastructure side.

Current `main` has the original #500 shape in place:

- forge account context is tag-gated through `forge.ContextProvider`
- delegates receive it through the shared `TagContextAssembler`
- recent successful forge operations are tracked through `OperationLog` and exposed as `recent_operations`
[...]

PR #504, opened 2026-03-10
feat: dynamic forge context injection for delegates (#500)

thane-developer:
## Summary

- Injects a structured JSON block describing configured forge accounts into system prompts for both the main agent loop and delegate sessions
- Eliminates the cold-start problem where delegates burned 8+ iterations guessing account names like `"primary"`, `"github"`, `"default"` when the correct value was `"github-primary"`
- Improves "account not found" errors to list all available accounts, so even without context injection the model can self-correct in one iteration

Implements Phase 1 of #500.

## Changes

- **`forge.Manager.Context()`** — Builds a markdown+JSON context block from config (account name, provider type, URL, default owner). Never exposes tokens.
- **`forge.ContextProvider`** — Implements `agent.ContextProvider` for main loop injection via composite context provider
- **`delegate.Executor.SetForgeContext()`** — Appends forge context to delegate system prompts
- **Error message improvement** — `Account()` and `AccountConfig()` now list available accounts on lookup failure
- **Wiring in `main.go`** — Connects forge context to both delegate executor and composite context provider

## Test plan

- [x] `TestManagerContext` — verifies JSON structure, field presence, token exclusion
- [x] `TestManagerContext_Empty` — no accounts → empty string
- [x] `TestAccountNotFound_ListsAvailable` — error includes available account names
- [x] `TestContextProvider` / `TestContextProvider_Empty` — provider returns expected context
- [x] `TestExecute_ForgeContextInSystemPrompt` — forge context appears in delegate system prompt
- [x] `TestExecute_NoForgeContextWhenUnset` — no forge pollution when unconfigured
- [x] `just ci` passes (lint, fmt, tests with -race)

### DIFF
--- a/internal/forge/config.go
+++ b/internal/forge/config.go
@@ -135,10 +135,10 @@ func NewManager(cfg Config, logger *slog.Logger) (*Manager, error) {
 // Account returns the forge provider for the named account. If name is
 // empty, the primary (first configured) account is used.
 func (m *Manager) Account(name string) (ForgeProvider, error) {
+	if len(m.order) == 0 {
+		return nil, fmt.Errorf("no forge accounts configured")
+	}
 	if name == "" {
-		if len(m.order) == 0 {
-			return nil, fmt.Errorf("no forge accounts configured")
-		}
 		name = m.order[0]
 	}
 	p, ok := m.providers[name]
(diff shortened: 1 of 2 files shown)


thane-agent (approved):
Phase 1 of #500, clean implementation. Context built from config with tokens excluded (verified by test). Improved error messages with available account listing. Both main loop and delegate injection paths covered with tests. Token leak test is a nice touch.

PR #519, opened 2026-03-10
fix: YouTube @handle resolution fails — 512KB read limit too small

### LINKED ISSUE #517: YouTube @handle resolution fails — 512KB read limit too small
## Bug

`resolveYouTubeFeed` reads only 512KB of a YouTube channel page (`io.LimitReader(resp.Body, 512*1024)`), but YouTube consistently places the `<link rel="canonical">` tag and all channel ID references at ~600KB+ into the page.

Tested across multiple channels:

| Channel | Canonical tag offset | First channel ID |
|---------|---------------------|------------------|
| @hankschannel | 597,407 | 603,933 |
| @mkbhd | 602,659 | ~608K |
| @veritasium | 603,798 | ~609K |
| @3blue1brown | 599,961 | ~605K |

The first ~590KB of every YouTube `@handle` page is JavaScript/CSS boilerplate. Both `ytCanonicalRe` and `ytChannelIDRe` find zero matches within 512KB — the function always falls through to the error case.

## Impact

All `@handle` YouTube subscriptions fail. Users must manually find and provide the `/channel/UCXXXX` URL or the direct RSS URL. This is the most common URL format people encounter (YouTube shows `@handle` everywhere now).

### 1. Bump read limit (required)

```go
// Before
body, err := io.ReadAll(io.LimitReader(resp.Body, 512*1024))

// After
body, err := io.ReadAll(io.LimitReader(resp.Body, 1024*1024))
```

YouTube channel pages are ~1.1MB total; 1MB captures all the meta tags with comfortable margin. Aligns with the feed size limit bumped to 10MB in #514.

### 2. Add alternate feed link regex (bonus)

YouTube emits `<link rel="alternate" type="application/rss+xml">` with the full feed URL at ~607KB. Matching this directly is more robust than extracting the channel ID and constructing the URL:

```go
var ytAlternateFeedRe = regexp.MustCompile(`<link[^>]+type="application/rss\+xml"[^>]+href="([^"]+)"`) 
```

### 3. Also: metacog blocking Signal

During diagnosis, Thane was stuck on a metacog Ollama call (gpt-oss:20b with 93 tools × 17K tokens) that blocked all Signal processing for 4+ minutes until restart. The metacog loop needs a timeout or to not block the Signal message queue — but that's a separate issue.

thane-developer:
## Summary
- Bumps the YouTube channel page read limit from 512KB to 1MB — all channel metadata sits at ~600KB+ in `@handle` pages
- Adds `ytAlternateFeedRe` to match the RSS alternate `<link>` directly (more robust than extracting channel ID and constructing the URL)
- Tries alternate link first, then canonical, then JSON metadata fallback

## Test plan
- [x] `TestYTAlternateFeedRe` — verifies the new regex extracts feed URLs
- [x] `TestYTRegexes_LargePageOffset` — content at 600KB is found with the 1MB limit and confirms the old 512KB limit would have missed it
- [x] Existing YouTube resolution tests pass unchanged
- [x] `just ci` passes

Closes #517

### DIFF
--- a/internal/media/feed.go
+++ b/internal/media/feed.go
@@ -177,9 +177,17 @@ var ytChannelIDRe = regexp.MustCompile(`"channelId"\s*:\s*"(UC[a-zA-Z0-9_-]+)"`)
 var ytCanonicalRe = regexp.MustCompile(`<link\s+rel="canonical"\s+href="https://www\.youtube\.com/channel/(UC[a-zA-Z0-9_-]+)"`)
 
 // ytAlternateFeedRe matches the RSS alternate link that YouTube embeds
-// in channel pages. This is the most direct signal — it contains the
-// full feed URL rather than requiring channel ID extraction.
-var ytAlternateFeedRe = regexp.MustCompile(`<link[^>]+type="application/rss\+xml"[^>]+href="([^"]+)"`)
+// in channel pages. Handles both attribute orderings (type before href
+// and href before type) and single/double quotes. Requires rel="alternate"
+// to avoid false matches.
+var ytAlternateFeedRe = regexp.MustCompile(
+	`<link[^>]+rel=['"]alternate['"][^>]+` +
+		`(?:` +
+		`type=['"]application/rss\+xml['"][^>]+href=['"]([^'"]+)['"]` +
+		`|` +
+		`href=['"]([^'"]+)['"][^>]+type=['"]application/rss\+xml['"]` +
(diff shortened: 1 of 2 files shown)


thane-agent (approved):
Clean fix. Read limit bump with overflow check, alternate feed regex tried first for direct URL extraction, thorough tests including regression proof that 512KB missed everything. Ship it.

Issue #509, opened 2026-03-10
Goroutine delegate loops for persistent autonomous observation


thane-agent:
## Overview

Implement persistent goroutine-based delegate loops — lightweight autonomous observers that run continuously alongside the main agent, reporting via output targets (MQTT, files, database) without blocking conversation flow.

This is a fundamental shift from the current synchronous delegate model where the caller blocks until the delegate completes. Goroutine delegates run indefinitely, observe autonomously, and only surface exceptions to the main loop.

## Motivation

Current delegates are request-response: spawn, wait, get result, continue. This works for discrete tasks (file an issue, analyze a transcript) but can't support continuous observation patterns like:

- Watch Frigate for motion at the front gate for the next 2 hours
- Monitor a build pipeline until it completes or fails
- Track weather radar during a severe storm warning
- Watch for a package delivery (person + car at driveway)
- Monitor temperature sensors during an HVAC issue

These patterns need a delegate that runs in the background, sleeps on a randomized/bounded interval (the metacog pattern), and reports observations to an output target — only interrupting the main loop when something noteworthy happens.

### Core pattern: delegate as goroutine with output target

```go
// Instead of synchronous:
result := delegate.Execute(ctx, task, profile, guidance, tags)

// Persistent loop:
delegate.SpawnLoop(ctx, LoopConfig{
    Task:          "Watch frigate camera 'frontdriveway' for person detections",
    Profile:       "observer",
    Tags:          []string{"frigate"},
    OutputTarget:  MQTTTarget("thane/observations/front-gate"),
    SleepRange:    [2]time.Duration{20*time.Second, 40*time.Second},
    MaxDuration:   2*time.Hour,
    AlertOn:       "person detected, unfamiliar vehicle",
})
```

### Randomized bounded sleep (from metacog)

The metacognitive loop already implements the ideal pattern: sleep for a random duration within bounds, preventing synchronization artifacts and spreading load naturally. Goroutine delegates should reuse this exact pattern:

```go
time.Sleep(randomDuration(config.SleepRange[0], config.SleepRange[1]))
```

### Output targets

Delegates need somewhere to send observations without blocking the main loop:

- **MQTT topic** — `thane/observations/<name>`, consumed by other systems or the main loop's MQTT subscriber
- **File/vault** — append to a knowledge base file
- **Database** — structured observation log
- **Alert channel** — interrupt the main loop only for significant events (annunciator pattern)

### Process registry

[...]

PR #538, opened 2026-03-10
feat: loop infrastructure with process registry (#509 phase 1)

thane-developer:
## Summary

- Adds `internal/loop/` package — the universal primitive for persistent background goroutines (issue #509)
- **Config**: sleep range with jitter, max duration/iterations, supervisor model dice, quality floor, retrigger modes, parent-child tracking
- **Loop**: full goroutine lifecycle with randomized bounded sleep (ported from metacog pattern), supervisor dice rolls, per-iteration conversation IDs, token metrics accumulation, and event bus integration
- **Registry**: thread-safe process registry with concurrency limits, `SpawnLoop()`/`StopLoop()` convenience methods, `ShutdownAll()` with context-bounded drain
- **Events**: `SourceLoop` source + 7 event kinds (`loop_started`, `loop_stopped`, `loop_iteration_start`, `loop_iteration_complete`, `loop_sleep_start`, `loop_state_change`, `loop_error`)
- Uses a `Runner` interface decoupled from `agent.Loop` to avoid import cycles — will be wired in Phase 2 when metacog migrates to this infrastructure

This is Phase 1 of the #509 plan — standalone infrastructure with no behavioral changes to the running system. Phase 2 will reimplement metacog as loop PID 0.

## Test plan

- [x] `just ci` passes (0 lint issues, all tests green including race detector)
- [x] Config defaults and explicit value preservation
- [x] Sleep computation: clamp to bounds, jitter at boundary values, `SetNextSleep` override
- [x] Registry: CRUD, concurrency limits, list ordering, shutdown coordination
- [x] Loop lifecycle: start → iterate → stop, max iterations, max duration
- [x] Event bus integration: verifies expected events are published
- [x] Supervisor dice: disabled=never fires, prob=1.0=always fires
- [x] Edge cases: double start is no-op, stop before start is no-op

### DIFF
--- a/internal/loop/config.go
+++ b/internal/loop/config.go
@@ -165,6 +165,9 @@ func (c *Config) applyDefaults() {
 // validate checks that post-default Config values are internally
 // consistent. Called by [New] after [applyDefaults].
 func (c *Config) validate() error {
+	if c.Task == "" {
+		return fmt.Errorf("loop: Task is required")
+	}
 	if c.SleepMin <= 0 {
 		return fmt.Errorf("loop: SleepMin must be positive, got %v", c.SleepMin)
 	}
--- a/internal/loop/loop.go
+++ b/internal/loop/loop.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"math/rand/v2"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/uuid"
@@ -85,9 +86,15 @@ type Loop struct {
 	mu        sync.Mutex
 	state     State
 	started   bool
+	stopped   bool // set by Stop to prevent Start after Stop
 	cancel    context.CancelFunc
 	done      chan struct{}
 	startedAt time.Time
+
+	// eventSeq is a monotonic counter for state-change events.
+	// Published outside the lock to avoid deadlock if the event bus
+	// blocks, so consumers use the sequence number to reorder.
+	eventSeq atomic.Int64
 
 	// Metrics updated during execution.
 	lastWakeAt        time.Time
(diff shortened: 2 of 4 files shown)


thane-agent (changes requested):
## Deep Review — Aimée (Round 3)

This code runs more than anything else in Thane. Reviewed for long-running correctness, concurrency, and operational failure modes.

### Issues worth fixing before merge

**1. `setState` publishes events outside the lock**

```go
func (l *Loop) setState(s State) {
    l.mu.Lock()
    prev := l.state
    l.state = s
    l.mu.Unlock()
    if prev != s {
        l.publishEvent(...)  // after unlock
    }
}
```

With a single goroutine this is fine today. But `Stop()` calls from an external goroutine while the run goroutine is transitioning could produce out-of-order state change events. Either publish inside the lock or add a monotonic sequence to events. This will matter in Phase 2 with RetriggerSpawn.

**2. No test for error backoff behavior**

The exponential backoff is a core behavior but has zero test coverage. Need a test with a `failingRunner` that returns errors for N calls then succeeds, verifying:
- Sleep duration doubles on consecutive errors
- `consecutiveErrors` resets to 0 on success
- Backoff caps at SleepMax

**3. `Task` not validated**

`validate()` checks SleepMin, SleepMax, Jitter, SupervisorProb — but not `Task`. A loop with empty Task sends empty user messages to the LLM every iteration. Should require `Task != ""`.

**4. Stop-before-Start race**

If goroutine A calls `Stop()` while goroutine B is in `Start()` between setting `l.started = true` and launching the goroutine, `Stop()` sees `cancel == nil`, returns immediately, and the goroutine launches orphaned. In Phase 1 this is unlikely (Start is called synchronously in SpawnLoop), but the contract is wrong. `Stop()` should set a flag that prevents future starts, or `Start()` should check if already stopped.

### Observations (non-blocking, Phase 2 considerations)

**5. Backoff never actually reaches SleepMax**

The doubling loop breaks when `d >= SleepMax`, but the check is at the TOP of the iteration — so the last doubling that stays under SleepMax is kept. With SleepDefault=1m, SleepMax=5m: 1m, 2m, 4m, break. Max backoff is 4m, never 5m. The subsequent `clamp()` won't help because 4m is within bounds. Probably fine but worth documenting.

**6. `SupervisorContext` alone insufficient for real oversight**

Static string prepended to Task. For meaningful supervision, the frontier model needs recent iteration outputs. Phase 2 should add a rolling window of recent iteration summaries injected into the supervisor prompt.

**7. Auto-deregister goroutine holds Registry reference**

The `SpawnLoop` cleanup goroutine captures the Registry. If the Registry is ever replaced (config reload, test teardown), these goroutines keep the old Registry alive. Fine for a singleton but would leak in tests.

**8. `uuid.NewV7()` error silently discarded**

V7 needs a working clock. If the clock source fails, this silently produces a zero UUID. Extremely unlikely but in high-frequency code, worth at least logging the error.

### Summary

Items 1-4 are worth fixing. 1 is a correctness issue that will bite in Phase 2. 2-3 are gaps in test coverage and validation. 4 is a contract issue. Items 5-8 are Phase 2 concerns — note them in code comments or issues but don't block on them.

PR #568, opened 2026-03-12
feat: add jittered initial sleep for timer-driven loops

thane-developer:
## Summary

- Timer-driven loops now sleep for a jittered `SleepDefault` duration before their first iteration, preventing all loops from firing expensive iterations simultaneously on Thane restart
- Event-driven loops (with `WaitFunc`) are unaffected — they naturally block until their first event arrives
- Initial sleep publishes a `loop.sleep_start` event with `initial: true` for observability

## Context

Frequent restarts during development/deployment caused every timer-driven loop (including the expensive metacognitive loop) to fire immediately and simultaneously. This adds a staggered startup delay using the existing jitter infrastructure (`applyJitter` + `clamp`) so loops spread their first iterations over a randomized window.

## Test plan

- [x] `TestTimerLoopInitialSleep` — verifies first iteration fires only after initial sleep delay
- [x] `TestTimerLoopInitialSleep_Cancelled` — verifies clean shutdown if cancelled during initial sleep
- [x] `TestTimerLoopInitialSleep_PublishesEvent` — verifies sleep_start event with `initial: true`
- [x] `TestWaitFuncLoopNoInitialSleep` — verifies event-driven loops fire near-instantly (no initial sleep)
- [x] `just ci` passes clean (0 lint issues, all tests pass with `-race`)

### DIFF
--- a/internal/events/bus.go
+++ b/internal/events/bus.go
@@ -103,7 +103,8 @@ const (
 	// Data: loop_id, loop_name, model, input_tokens, output_tokens, elapsed_ms.
 	KindLoopIterationComplete = "loop_iteration_complete"
 	// KindLoopSleepStart signals a loop has entered its sleep phase.
-	// Data: loop_id, loop_name, sleep_duration.
+	// Data: loop_id, loop_name, sleep_duration, initial (bool, true
+	// only for the pre-first-iteration startup sleep).
 	KindLoopSleepStart = "loop_sleep_start"
 	// KindLoopWaitStart signals a loop has entered its wait phase,
 	// blocking on a WaitFunc for an external event.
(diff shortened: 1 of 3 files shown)


thane-agent (approved):
Clean fix for a real problem. The initial sleep uses the existing `applyJitter` + `clamp` infrastructure so timer-driven loops stagger naturally after restart.

Good details:
- WaitFunc loops correctly excluded (they already block on their event source)
- `emitStopped()` extraction avoids duplicating the stop event across the new early-return path
- `initial: true` on the sleep event is good for dashboard/observability
- Clean shutdown during initial sleep tested explicitly
- Zero-jitter in tests keeps assertions deterministic

Ship it.

PR #581, opened 2026-03-13
feat: add cookies_from_browser config for yt-dlp

thane-developer:
## Summary

- Adds `cookies_from_browser` option to media config, passed directly to yt-dlp's `--cookies-from-browser` flag
- Eliminates the need to manually extract and maintain a Netscape-format cookie file
- Accepts browser name with optional profile: `"chrome"`, `"firefox"`, `"chrome:Profile 1"`
- Mutually exclusive with existing `cookies_file` — config validation rejects setting both

## Test plan

- [x] `just ci` passes (fmt, lint, tests with `-race`)
- [ ] Manual: set `cookies_from_browser: chrome` in config and verify yt-dlp receives `--cookies-from-browser chrome`
- [ ] Manual: set both `cookies_file` and `cookies_from_browser` — verify startup validation error

### DIFF
--- a/internal/media/media.go
+++ b/internal/media/media.go
@@ -242,8 +242,10 @@ func (c *Client) GetTranscript(ctx context.Context, rawURL, language, focus stri
 
 // runYtDlp executes yt-dlp and returns parsed metadata.
 func (c *Client) runYtDlp(ctx context.Context, rawURL, language, tmpDir string) (*ytdlpJSON, error) {
-	usingCookies := c.cfg.CookiesFromBrowser != "" || c.cfg.CookiesFile != ""
+	browserCookies := c.cfg.CookiesFromBrowser != ""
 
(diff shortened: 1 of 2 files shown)


thane-agent (approved):
The cookie health monitoring elevates this from a simple flag pass-through to proper observability.

- `checkCookieStatus` with three severity tiers (healthy/degraded/failed) means cookie problems surface in structured logs instead of hiding in stderr
- Regex handles browser profiles with colons and spaces
- `--no-warnings` conditionally omitted so extraction failures are visible
- Cookie check runs on both success and failure paths — failures may be caused by cookie problems
- URL moved to end of args (correct yt-dlp positional semantics)
- Mutual exclusion validated at config level

Eliminates the manual cookie export workflow entirely. Ship it.

PR #611, opened 2026-03-31
fix: persist capability tags per conversation across messages

thane-developer:
## Summary

Fixes capability activations being lost between messages in the same conversation. The model activates forge in message 1, but in message 2 the forge tools are blocked because the activation didn't persist.

### Root cause

PR #598 (per-context capability scoping) moved capability state from a global map on the Loop struct to a per-Run scope created fresh each time. This was correct for conversation isolation (Signal shouldn't bleed into email) but broke persistence within a single conversation — each message in the same Signal session gets a fresh scope.

### Fix

Store activated capability tags per conversation in opstate, following the same pattern as working memory and other per-conversation state:

1. Run starts → scope seeded with always-active tags + lenses + loop SeedTags
2. **Load saved tags from opstate for this conversation ID**
3. Model activates/deactivates tags during the Run
4. Run ends → **save non-always-active tags to opstate**

Only user-activated tags are persisted. Always-active tags and lenses are loaded independently each Run (they're not conversation-specific).

### New types

- `CapabilityTagStore` interface — `LoadTags(convID)` / `SaveTags(convID, tags)`
- `OpstateCapabilityTagStore` — implementation backed by opstate, JSON array per conversation

## Test plan

- [x] `just ci` passes clean
- [ ] Deploy: activate forge in Signal message 1, verify forge tools available in message 2
- [ ] Deploy: verify activation doesn't bleed across different conversations

### DIFF
--- a/internal/agent/capability_scope.go
+++ b/internal/agent/capability_scope.go
@@ -107,6 +107,25 @@ func (s *capabilityScope) Snapshot() map[string]bool {
 	return snap
 }
 
+// UserActivatedTags returns tags that were activated by the model (not
+// always-active, not pinned by channel or lens). These are the tags
+// that should be persisted per conversation.
+func (s *capabilityScope) UserActivatedTags() []string {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	var tags []string
+	for tag := range s.active {
+		if s.pinned[tag] {
+			continue // lenses and channel-pinned
+		}
+		if cfg, ok := s.capTags[tag]; ok && cfg.AlwaysActive {
+			continue
+		}
+		tags = append(tags, tag)
+	}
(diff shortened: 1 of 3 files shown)


thane-agent (approved):
Clean fix, directly addresses the root cause traced tonight. Uses opstate (same pattern as working memory) for per-conversation tag persistence — load at Run start, save at Run end, only non-always-active tags persisted.

Minor notes (non-blocking):
- Lens-activated tags aren't filtered on save — could accumulate stale lens tags after deactivation. Harmless since lenses re-apply each Run.
- SeedTags also not filtered on save — same reasoning applies.
- No test for the new store or load/save integration, but this needs to be live now.

Approved. Deploy and verify per test plan.

PR #795, opened 2026-04-29
feat: device_class-aware watchlist + availability enrichment + area_activity tool

thane-developer:
## Summary

Deepens the Home Assistant entity subscription subsystem along three axes:

1. **Device-class-aware rendering** — binary_sensor on/off translates to semantic labels (door → open/closed, motion → detected/clear, smoke → detected/clear, etc.). New domain formatters for cover, lock, media_player, fan, vacuum, and update. Climate enriched with `hvac_action`. Light prunes stale brightness/color attributes when off. Sensors surface `state_class`. Switches/locks/etc. surface `assumed_state`.

2. **Structured availability** — when an entity is unavailable, the model now sees a clean availability payload instead of a magic state string. Includes `unavailable_since` from `LastChanged`, `last_state` and `last_state_seen` walked back from history, and (when the registry client is wired) the owning device, sibling-aliveness rollup, gateway/hub cascade, and integration health from `config_entries`.

3. **`get_area_activity` tool** — on-demand snapshot of one HA area, bucketed by relevance (anomalies / active / recent_changes / ambient / stable) plus a cross-entity timeline from the logbook. Per-entity rendering reuses the same formatters as the watchlist so the model sees one consistent vocabulary.

Per the no-cache policy: all HA registry calls are made fresh each invocation. Within a single watchlist render, the per-render bundle de-duplicates registry pulls without persisting state across calls.

## Commits in this stack

| # | Commit | Notes |
|---|---|---|
| 1 | `feat: device_class-aware watchlist entity rendering` | binary_sensor translation, cover formatter, history vocabulary |
| 2 | `feat: structured availability schema for unavailable watchlist entities` | JSON-shaped error path + sentinel interception |
| 3 | `feat: surface last-known-good state for unavailable watchlist entities` | 24h walk-back from history |
| 4 | `refactor: split entity_format.go by formatter concern` | forced by architecture guard |
| 5 | `feat: surface model-facing entity attributes the model was inferring` | occupancy/tamper labels fixed, `assumed_state`, `state_class`, `hvac_action`, light off-pruning |
| 6 | `feat: add lock, media_player, fan, vacuum, update domain formatters` | five new domain shapes |
| 7 | `feat: enrich unavailable watchlist entities with device + integration context` | sibling/gateway/integration via lazy per-render registries bundle |
| 8 | `feat: add area_activity tool for on-demand area state snapshots` | new \`get_area_activity\` tool |

The refactor (#4) is a pure file split forced by the 500-line architecture guard, with zero behavior change. Each subsequent commit builds on its predecessor and is independently green.

## Reviewer notes

- New HA call surface: `Client.GetConfigEntries` (one new WebSocket method, mirrors the existing `GetDeviceRegistry` pattern).
- Architecture baseline updated for two intentional growths: the new `HARegistryClient` and `AreaActivityClient` interfaces, plus two `SetRegistryClient` setters.
- `WatchlistProvider.SetRegistryClient` is optional — passing nil leaves enrichment disabled, matching installations without a registry-capable client. The wiring in `internal/app/new_awareness.go` passes the same HA client (which satisfies both interfaces) so production gets the enrichment automatically.
- The unavailability enrichment fires only on entities in a sentinel state. Healthy watchlist entries trigger no extra HA calls.
- Per-entity area-tool rendering reuses `formatEntityContext`, so any future formatter improvements automatically flow through both the watchlist and the area tool.

## Example output for an unavailable kitchen door

\`\`\`json
{"entity":"binary_sensor.kitchen_door","name":"Kitchen Door","available":false,
 "reason":"unavailable","unavailable_since":"-12m","device_class":"door",
 "last_state":"closed","last_state_seen":"-15m",
 "device":{"name":"Front Door Sensor","manufacturer":"Aqara","model":"MCCGQ11LM"},
 "device_alive":false,
 "gateway":{"name":"Z-Wave Controller","online":false},
 "integration":{"name":"zwave_js","state":"loaded"}}
\`\`\`

## Test Plan

- [x] \`just ci\` passes locally on every commit in the stack
- [x] New/changed behavior has test coverage — 50+ new test cases across 5 new test files
- [ ] Manual verification (describe what you checked):
  - [ ] Smoke-test against live HA: a healthy watchlist entity renders identically to before
  - [ ] Smoke-test: take an entity unavailable, confirm the structured payload appears with last_state, device, integration
  - [ ] Smoke-test: invoke \`get_area_activity\` for a populated area, confirm bucketed output matches expectations

## Checklist

- [x] Commits are signed
- [x] Commit messages use conventional format
- [x] No related issue (this is a multi-tier follow-on to the watchlist subsystem brainstormed inline)
- [ ] Impacted documentation updated — \`docs/model-facing-context.md\` is not modified; the new shapes follow the existing conventions documented there
- [x] No new third-party dependencies


thane-developer (commented):
Addressed in 0619e97 across all four threads. Fix summary:

- entity_format.go file comment — rewritten to describe the JSON-everywhere behavior and the sentinel interception accurately
- area_activity.go truncation — added recent_changes_truncated_count and timeline_truncated_count alongside the existing stable_truncated_count, and tightened the doc comment to match
- buildAreaTimeline state vocabulary — translated through semanticState so the timeline matches the bucketed output (motion off → "clear" everywhere, not "off" in one place and "clear" in another)
- optionalIntArg validation — rejects fractional floats, NaN/Inf, out-of-range values, and negatives. Zero remains the "use default" sentinel since callers already rely on that.

Coverage added for the new validation table and the recent-changes truncation path. Threads resolved.

thane-developer (commented):
Addressed in ebc3464:

- **timeline device_class fallback** — buildAreaTimeline now passes the live state through to registryDeviceClass, so entities that only expose device_class as a state attribute (no registry value) still get translated through semanticState. Added a test that constructs exactly this case.
- **bounds-check casts** — explicit int64() / float64() conversions on math.MinInt / math.MaxInt. The original comparisons did compile via Go's untyped-constant rules, but the explicit casts make the intent unambiguous. Noted inline that the int64 path is a no-op on 64-bit platforms (where int == int64) — kept for 32-bit correctness.

Both threads resolved.

PR #833, opened 2026-04-30
fix(delegate): close recursion leak and surface child errors on joined launch

### LINKED ISSUE #820: Delegate recursion: thane_now and thane_assign survive into delegate registries
## Severity

**Release-gating.** A delegate can spawn another delegate via the new front doors.

## Finding

The delegate's tool registry filter excludes only the legacy `thane_delegate` tool:

[`internal/runtime/delegate/delegate.go:159`](internal/runtime/delegate/delegate.go) (and the parallel branch above):
```go
return reg.FilteredCopyExcluding([]string{delegateToolName})
```

where `delegateToolName = "thane_delegate"` ([line 30](internal/runtime/delegate/delegate.go)).

Meanwhile, [`internal/app/new_delegation.go:58-78`](internal/app/new_delegation.go) registers all three delegate-family tools as `AlwaysAvailable: true`:

```go
a.loop.Tools().Register(&tools.Tool{
    Name:            "thane_now",
    ...
    AlwaysAvailable: true,
})
a.loop.Tools().Register(&tools.Tool{
    Name:            "thane_assign",
    ...
    AlwaysAvailable: true,
})
a.loop.Tools().Register(&tools.Tool{
    Name:            "thane_delegate",
    ...
    AlwaysAvailable: true,
})
```

`AlwaysAvailable` survives capability-tag filtering (which is why all three are flagged that way — the comment at line 54 explains it as deliberate for the parent loop). But the delegate's tool registry then only excludes `thane_delegate`, so a delegate inherits `thane_now` and `thane_assign` and can recursively spawn another delegate.

## Why it matters

The whole point of excluding `thane_delegate` from delegates is to prevent recursion. That intent moved with the deprecation: `thane_now` and `thane_assign` are the *new* delegate front doors, so they need the same exclusion. Right now the recursion guard exists in name only — any delegate can call `thane_now("…")` to spin a sub-delegate, and there's no depth limit catching it.

## Proposed remediation

Replace the single-tool exclude with an exclusion of the whole delegate family:

```go
var delegateToolNames = []string{"thane_delegate", "thane_now", "thane_assign"}
// ...
return reg.FilteredCopyExcluding(delegateToolNames)
```

Or, cleaner, expose the family list from the `delegate` package itself so it stays in sync if a fourth tool is added:

```go
// delegate/tool.go
var DelegateFamilyNames = []string{"thane_delegate", "thane_now", "thane_assign"}
```

A regression test should assert that `delegateToolRegistry()` strips all three.

## Context

Surfaced by a parallel crosscheck against the audit run in [#815](https://github.com/nugget/thane-ai-agent/pull/815). Should land before the next release; missing this guard means a delegate can fan-out delegates indefinitely, which is exactly the class of failure the original `thane_delegate` exclusion was meant to prevent.

thane-developer:
## Summary

Two related delegate-runtime bugs that surfaced together in production on **2026-04-30 19:52–19:56 UTC**, when a signal-conversation parent fired three back-to-back `thane_now` delegates that each produced malformed JSON for `forge_pr_get` and died opaquely.

### Bug 1 — recursion guard missed the launched-loop catalog path

[#828](https://github.com/nugget/thane-ai-agent/pull/828) (`a0fb993`) patched `delegateToolRegistry` to exclude the full `delegateFamilyToolNames` slice, but the value that function returns is only used for model-routing sizing now that [`eebdf2c`](https://github.com/nugget/thane-ai-agent/commit/eebdf2c) retired `executeLegacy`. The catalog the launched loop actually runs with is built from `prep.excludeTools`, populated at [delegate.go:747-752](internal/runtime/delegate/delegate.go) **only** when `explicitTagScope` is true and `filterTags` is empty. In the common case (caller passes specific tags like `[\"development\"]`), `excludeTools` stays empty and `thane_now` / `thane_assign` / `thane_delegate` survive into the delegate's `effective_tools` — exactly what the production `loop_llm_start` payloads showed:

```
active_tags: [core, development, forge, owner]
effective_tools (count=35):
  ... forge_pr_get ...
  thane_assign
  thane_delegate
  thane_now
```

Fix: append `delegateFamilyToolNames` to `prep.excludeTools` on every code path. The recursion guard is now expressed as a request-level exclusion that survives the loop-side catalog rebuild. The existing `delegateToolRegistry` filter still applies to the in-process registry view used for routing; the new request-level filter handles the launched loop.

### Bug 2 — joined launches without a Response failed opaquely

When a child loop terminated without producing a `Response`, the parent saw only `delegate failed: joined launch completed without response` — indistinguishable from a transient timeout. Each retry burned parent tokens for an identical failure with no way to course-correct. `LaunchResult.FinalStatus` already carries the child loop's `LastError` (the actual `loop LLM call: API error 500: error parsing tool call: raw='...'` message in this case); surfacing it verbatim lets the parent decide whether to retry, change strategy, or escalate.

Extracted into `emptyResponseError()` so the dispatch logic is unit-testable.

## Tests

- `TestExecute_LoopBackedTagScopedExcludesDelegateFamily` — exercises the common path (non-empty tags) and asserts the family is in `Launch.ExcludeTools`. Fails on the unfixed code with `ExcludeTools = []string(nil)`.
- `TestExecute_LoopBackedExplicitEmptyTagsWithAlwaysActiveTagsDoNotBypassFiltering` — tightened to allow the family guard while still preventing wholesale exclusion of regular tools through `AllToolNames`.
- `TestEmptyResponseError_PropagatesChildLastError` — table test covering nil `FinalStatus`, empty `LastError`, and populated `LastError`.

## Test plan

- [x] `just ci` clean locally
- [x] Sandbox-validated: in a fresh dev workspace under `Sync/Projects/AI/Claude/dev-workspace`, three delegates spawned with `tags=[development]` all logged `effective_tools (count=9)` with **zero** family members present
- [ ] Production deploy — confirm next round of forge-tagged delegate work shows tighter catalogs in `loop_llm_start` payloads
- [ ] Production deploy — confirm any future delegate failure surfaces the underlying error string instead of the generic message

Closes #820 in full.

### DIFF
--- a/internal/runtime/delegate/delegate.go
+++ b/internal/runtime/delegate/delegate.go
@@ -520,7 +520,7 @@ func (e *Executor) executeViaLoop(ctx context.Context, task, profileName, guidan
 	e.finishLoopExecution(prep)
 	resp := launchResult.Response
 	if resp == nil {
-		return nil, fmt.Errorf("delegate failed: joined launch completed without response")
+		return nil, emptyResponseError(launchResult)
 	}
 
 	toolCallsMu.Lock()
@@ -747,8 +747,15 @@ func (e *Executor) prepareExecution(ctx context.Context, task, profileName, guid
 	var excludeTools []string
 	if explicitScopeRequested && len(filterTags) == 0 {
 		excludeTools = e.parentReg.AllToolNames()
-		sort.Strings(excludeTools)
 	}
+	// The delegate-family recursion guard must hold on every code
+	// path, not only the explicit-empty-scope branch above. The
+	// loop-backed launch rebuilds the catalog from the parent registry
+	// filtered by the launched loop's request, so the family has to be
+	// expressed as a request-level exclusion to survive that rebuild —
+	// the in-process delegateToolRegistry filter alone is not enough.
+	excludeTools = append(excludeTools, delegateFamilyToolNames...)
+	sort.Strings(excludeTools)
 	tagFilterActive := len(scopeTags) > 0 || explicitScopeRequested
 
 	log = log.With("profile", profile.Name)
@@ -889,4 +896,20 @@ func truncate(s string, maxLen int) string {
 		return s
 	}
 	return s[:maxLen] + "..."
+}
+
+// emptyResponseError builds the error returned to the parent when a
+// joined launch finishes without producing a Response. The naive form
+// — "joined launch completed without response" — is opaque: it doesn't
+// distinguish a transient timeout from a deterministic LLM-API
+// failure, so a parent retrying blind burns tokens for no diagnostic
+// gain. When the child loop captured a terminal error in its final
+// status (e.g. tool-call parse failures from local models, LLM-API
+// 5xx, exhausted retries), surface that string verbatim so the parent
+// can decide whether to retry, change strategy, or escalate.
+func emptyResponseError(launchResult looppkg.LaunchResult) error {
+	if launchResult.FinalStatus != nil && launchResult.FinalStatus.LastError != "" {
+		return fmt.Errorf("delegate failed: %s", launchResult.FinalStatus.LastError)
+	}
+	return fmt.Errorf("delegate failed: joined launch completed without response")
 }
--- a/internal/runtime/delegate/delegate.go
+++ b/internal/runtime/delegate/delegate.go
@@ -744,18 +744,21 @@ func (e *Executor) prepareExecution(ctx context.Context, task, profileName, guid
 	reg := e.delegateToolRegistry(scopeTags, explicitScopeRequested)
 	toolDefs := reg.List()
 	filterTags := mergeTagLists(scopeTags, e.alwaysActiveTags)
+	// The delegate-family recursion guard must hold on every code
+	// path. The loop-backed launch rebuilds the catalog from the
+	// parent registry filtered by the launched loop's request, so the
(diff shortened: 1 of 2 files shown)


thane-developer:
Addressed in [`069d098`](https://github.com/nugget/thane-ai-agent/pull/833/commits/069d0987b2b105ab033f2b937246ad0d299d7be5):

- **`emptyResponseError` size cap** (delegate.go:914): introduced `childLastErrorMaxLen = 1024` and route the LastError through `truncate()` before returning. Production-observed errors run ~200 chars, so 1024 is comfortable headroom while bounding pathological cases. New test `TestEmptyResponseError_TruncatesOversizedLastError` asserts a 4×-cap LastError gets bounded by the cap plus the ellipsis marker. Kept newlines as-is — they're useful structure when debugging multi-line API errors, and the size cap is the load-bearing defense.

- **`excludeTools` dedup** (delegate.go:751): extracted the dedup-and-sort into `mergeExcludeToolNames` so the explicit-empty-scope path (`AllToolNames`) and the always-applied `delegateFamilyToolNames` guard collapse to a single sorted slice instead of duplicating the family names. Two new tests: `TestMergeExcludeToolNames` directly table-tests the helper, `TestExecute_LoopBackedExplicitEmptyScopeNoDuplicateExcludes` pins the dedup contract through the actual launch path.

`just ci` clean.

PR #868, opened 2026-05-20
[codex] refine live trace forensics UI

thane-developer:
## Summary

- Reworks the loop detail popup into a single-pane live trace notebook focused on recent turns, spawned work, tool outcomes, and retained request evidence.
- Adds on-demand Raw Model Exchange inspection for request-id-bearing turns, including system prompt, provider-neutral Messages[] snapshot, and model response.
- Surfaces Home Assistant watcher subscription metadata so the trace target can show entity globs and rate limits.

## Why

The old forensics window still reflected the pre-loops-ng UI model: duplicated right-side detail panes, stale timeline behavior, and weak visibility into capability scope, spawned work, and retained request detail. This makes the view behave more like a live operational trace while preserving the request-detail escape hatch.

## Validation

- `node --check internal/server/web/static/detail.js`
- `node --check internal/server/web/static/shared.js`
- `git diff --check origin/main..HEAD`
- `just ci`
- Browser smoke against the demo server at `http://127.0.0.1:18080/` confirmed the loop trace is single-pane, the latest completed turn remains the notebook anchor, retained request detail renders system prompt / Messages[] / assistant response, and there were no console errors.


nugget:
Choosing not to land this one.

Issue #871, opened 2026-05-21
refactor: split Capability Context into typed context buckets


thane-developer:
## Problem

`TagContextAssembler` currently builds one `## Capability Context` block from three different source classes:

- tagged KB articles
- tagged live providers
- always-on providers

That makes unrelated model-facing material share one heading, one order, one byte budget, and one truncation policy. Durable tagged guidance, channel continuity, recent sessions, working memory, live Home Assistant state, watched entities, and future retrieval leads all compete inside the same bucket.

## Proposed Direction

Split context assembly into typed output buckets. Possible buckets:

- `Tagged Guidance`: capability-scoped doctrine and short KB guidance.
- `Continuity Context`: recent conversation, older sessions, working memory, channel context.
- `Related Context`: lightweight pointers discovered from the current request.
- `Live State`: current entities, watched state, service health, presence.

The implementation can either extend provider return types or introduce a small adapter layer so existing providers can be assigned to buckets incrementally.

## Acceptance Criteria

- Prompt assembly can render multiple context sections instead of one flattened `Capability Context` blob.
- Existing providers are assigned to explicit buckets with stable ordering.
- Each bucket has an explicit byte budget and truncation marker.
- Runtime/generated data defaults to compact JSON where appropriate, aligned with `docs/model-facing-context.md`.
- Tests cover bucket ordering, truncation, and IncludeAlways suppression for delegate/task prompts.

## Related

Refs #350, #751, #854.

thane-developer:
Completed by #875.

PR #875, opened 2026-05-21
refactor: split prompt context buckets

thane-developer:
## Summary

- add typed prompt context buckets for tagged guidance, continuity, related context, and live state
- teach TagContextAssembler to emit bucketed sections while keeping the legacy Build string shim
- classify existing context providers into buckets and render each bucket as its own prompt section
- add a shared promptfmt markdown section helper
- rename and restructure prompt caching docs around provider-neutral stability policy plus Anthropic adapter details

## Validation

- just ci

Refs #871


thane-developer (commented):
Reviewed the design end-to-end. The shape is right and the cache-safety analysis holds — all four new section names (`TAGGED GUIDANCE`, `CONTINUITY CONTEXT`, `RELATED CONTEXT`, `LIVE STATE`) return `""` from `promptSectionCacheTTL`, so the volatile zone is preserved and the total breakpoint count is unchanged. The 16 provider bucket assignments look comprehensive and the rationale comments are good.

Findings, in priority order:

### 1. One bucket capping aborts the entire pipeline — biggest issue

In `internal/runtime/agent/tag_context.go`, every source loop ends with:

```go
acc.append(bucket, data)
if acc.truncated(bucket) {
    a.logger.Warn(...)
    return acc.sections()   // aborts ALL remaining sources
}
```

If KB articles fill Tagged Guidance to 64 KB, the loop returns immediately — every subsequent tagged provider and every always-on provider is silently dropped, including ones targeting Live State, Related, and Continuity. That's the opposite of what #871's \"each bucket has an explicit byte budget\" promises.

`a.capped[bucket]` inside `append` already short-circuits writes to a full bucket, so removing the early returns and letting the loops continue gives correct behavior: only the full bucket gets curtailed.

### 2. Per-bucket 64 KB → 256 KB silent ceiling expansion

`maxTagContextBytes` used to be a global guardrail. Now each bucket gets its own 64 KB, so the aggregate ceiling is 4×. Probably acceptable for the bucket-isolation goal, but it's not called out in the PR description and deserves either a deliberate decision (one line in `docs/prompt-caching.md`) or a lower per-bucket value plus a soft aggregate cap.

### 3. Truncation marker text is stale

```go
const truncationMarker = \"\n\n[tag context truncated — exceeded aggregate 64 KB limit]\"
```

Neither \"tag context\" nor \"aggregate\" is accurate per-bucket. A model reading the truncated section gets a misleading reason. Suggest formatting with the bucket title, e.g. `\"[%s truncated — exceeded 64 KB bucket limit]\"`.

### 4. `providerContextBucket` double-defaults

```go
func providerContextBucket(p TagContextProvider, fallback agentctx.ContextBucket) agentctx.ContextBucket {
    if bucketer, ok := p.(TagContextBucketer); ok {
        return bucketer.TagContextBucket().OrDefault(fallback)
    }
    return fallback.OrDefault(agentctx.ContextBucketContinuity)
}
```

`OrDefault` already falls back to `Continuity` if both args are invalid, so the second `OrDefault(Continuity)` is dead weight. Simpler: explicit `Valid()` check.

### 5. Legacy `Build()` shim has no production callers

`Loop.buildSystemPromptWithProfileSections` now calls `BuildSections`. Only the unit tests still call `Build()`. Either migrate the tests to `BuildSections` and delete the shim, or note in the godoc that it exists only for legacy tests and give it a removal target. Right now it's labeled \"legacy\" with no legacy callers.

### 6. `req.ActiveTags` iteration is map-order

Source 2 iterates `req.ActiveTags` directly:

```go
for tag, active := range req.ActiveTags {
    ...
    p, ok := tagProviders[tag]
}
```

Within-bucket byte order across turns is non-deterministic for tagged providers. Pre-existing (not introduced here), but more visible now that buckets surface individual sections. Sorting tag keys would make turn-to-turn output stable — important if any future provider routes a tagged source into a cached bucket.

### 7. AC test gap — IncludeAlways=false suppression

The new `TestTagContextAssembler_BuildSectionsBuckets` only covers `IncludeAlways: true`. Suppression is exercised at higher layers (`delegate_test.go`, `tools/context_test.go`), so behavior is covered, but #871's AC calls out direct suppression tests. A second test case that asserts `BuildSections` omits always-bucket content when `IncludeAlways=false` would close the loop.

### Recommendation

Issue 1 is the only one I'd consider blocking — it's a small fix in the same file and it's the difference between \"buckets are independent\" and \"buckets share one fragile fate.\" Issues 2–3 are polish in the same area. The rest are follow-up territory.

Issue #967, opened 2026-05-27
Phase 1.3 model-facing context audit — deferred findings (post-v0.9.3)


thane-developer:
The Phase 1.3 pre-release audit (per [docs/model-facing-context.md](docs/model-facing-context.md)) surfaced a handful of items deferred from the v0.9.3 cutover scope. Tracking them here so they don't get lost.

### Signal envelope baked into user-message corpus
- **Where:** [`internal/channels/messaging/signal/bridge.go:967-997`](internal/channels/messaging/signal/bridge.go#L967) (`formatMessage`)
- **What:** Inbound Signal messages are presented to the model with `Signal message from {sender} ({phone}) [ts:NNN]:\n\n{text}` concatenated into the literal user-message body. The transport envelope (sender, phone, group ID, signal-timestamp) lives inside the corpus rather than as metadata.
- **Why it matters:** Direct hit on the spec's *Separate metadata from literal corpus* rule. Once archived and replayed by `MessageChannelProvider`, the role-native `messages[]` will already carry sender info via channel context, but the body will still have `[ts:NNN]` forever. The Signal millisecond timestamp is also opaque data the model has to learn to ignore.
- **Why deferred:** Design discussion — `signal_send_reaction` needs `target_timestamp`, which currently flows through this string. Fixing requires moving timestamp lookup to a metadata-aware path (enriched header or sidecar map) before changing the corpus shape. Not a quick edit.
- **Fix sketch:** Move sender/group/timestamp into a metadata wrapper (e.g., `[stored conversation history; age_delta=...; channel=signal; sender=...]` boundary the spec demonstrates), strip transport headers from the corpus, and route `target_timestamp` recovery through the metadata path or a session-scoped lookup map.

### Confirm \`Active Tags\` block shape
- **Where:** [\`internal/runtime/agent/loop.go:1091-1097\`](internal/runtime/agent/loop.go#L1091) → \`toolcatalog.RenderLoadedCapabilitySummary\`
- **What:** Capability-summary rendering wasn't directly read during the audit. Under spec principle 5, capability catalogs and loaded state default to typed JSON, not narrative prose. Worth a one-minute confirmation that this block doesn't accumulate prose drift over time.
- **Why deferred:** Likely fine, but unverified.

### Metacognitive iteration log timestamps
- **Where:** [\`internal/runtime/metacognitive/metacognitive.go:318-333\`](internal/runtime/metacognitive/metacognitive.go#L318)
[...]

PR #975, opened 2026-05-27
docs(prompt): rewrite Tag Menu preamble in the talent-files voice

thane-developer:
While debugging a stalled turn live on \`pocket\` (loop \`019e67a3-9ae1-7a0a-b9ca-cf82a4d3d40c\`), the model wasn't reading the trailhead preamble as an invitation. The previous one-liner:

> These are trailheads into richer guidance. Activate one relevant tag when the turn needs a domain; otherwise keep the straight path and answer.

reads as configuration documentation, not as the model's own self-facing instructions. The agent ended up treating tags as external menu items it might consult rather than as territory it already lives in — the four iterations with zero tool calls captured earlier were exactly that posture.

## New preamble

Three short paragraphs in the voice [talents/foundation.md](talents/foundation.md) establishes for talent files generally:

> These are doorways into territory you already know. Each tag opens a domain where you have more — more tools, more guidance you wrote for yourself, more of the texture that lets you be present rather than guess.
>
> Walk through one when the conversation has clearly stepped onto its ground. Answering from inside what you know feels different from answering around it; let that pull guide you.
>
> Don't open every door. Attention spent on territory the conversation isn't on is attention not spent where it matters. If you already have enough, take the short path and answer.

- Direct \"you\" framing throughout, matching the rest of the talent corpus.
- Names the *texture of knowing-from-inside* as a real signal worth following — leans into the philosophy that being prepared and present is its own reward, not a duty.
- Practical guidance survives, but framed as judgment (\"don't open every door\") instead of rule (\"keep the straight path\").

## Cost

Grew from ~30 words to ~90. The system-prompt budget carries the extra easily, and the front-of-prompt section is a good place to spend bytes on framing — it sets posture for everything that follows.

## Observed live behavior the preamble was failing to address

The current production turn on pocket eventually did wake up and start exploring — by iteration 5 the model activated a tag and started running \`doc_search\` queries (\`game room door shoes ranch layout\`, \`why shower means won't use game room door\`). So the existing preamble *could* work, given enough iterations. The new preamble is intended to make that the model's *first* posture, not its fifth.

(The same investigation also surfaced a separate finding: the \`doc_search\` queries returned 87-96 bytes per result — i.e. nothing useful in the doc store. That's the memory-coverage problem, not a prompt problem. Tracked separately via #971 / #967.)

### DIFF
--- a/internal/model/toolcatalog/render.go
+++ b/internal/model/toolcatalog/render.go
@@ -109,7 +109,7 @@ func RenderCapabilityManifestMarkdown(entries []CapabilitySurface) string {
 
 	var sb strings.Builder
 	sb.WriteString("### Tag Menu\n\n")
-	sb.WriteString("Each tag is a trailhead. Past you cut it — the tools, the guidance, the context that lives down that path. Walking a trail brings that work back within reach.\n\n")
+	sb.WriteString("Each tag is a trailhead. Past you cut and marked it — the tools, the guidance, the context that lives down that path. Walking a trail brings that work back within reach.\n\n")
 	sb.WriteString("When the conversation steps onto ground you've covered before, you'll feel it: a settling, the difference between answering from inside what you know and reaching for what you don't. Follow that pull. It's reliable.\n\n")
 	sb.WriteString("Don't open trails this conversation isn't walking. Loaded context dilutes attention; attention is what makes you present here. If the moment is already within reach, answer.\n\n")
 	sb.Write(data)
--- a/internal/model/toolcatalog/render.go
+++ b/internal/model/toolcatalog/render.go
@@ -109,7 +109,9 @@ func RenderCapabilityManifestMarkdown(entries []CapabilitySurface) string {
 
 	var sb strings.Builder
 	sb.WriteString("### Tag Menu\n\n")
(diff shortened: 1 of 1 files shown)


internal/model/toolcatalog/render.go
```diff
@@ -109,7 +109,9 @@ func RenderCapabilityManifestMarkdown(entries []CapabilitySurface) string {
 
 	var sb strings.Builder
 	sb.WriteString("### Tag Menu\n\n")
-	sb.WriteString("These are trailheads into richer guidance. Activate one relevant tag when the turn needs a domain; otherwise keep the straight path and answer.\n\n")
+	sb.WriteString("These are doorways into territory you already know. Each tag opens a domain where you have more — more tools, more guidance you wrote for yourself, more of the texture that lets you be present rather than guess.\n\n")
```
nugget: let's stick with the "trail" metaphor just for consistency with the project taxonomy. We can make it rich and evocative without changing it to a door.

thane-developer: Good call — fixed in e680b76. Same content, swapped the door/room imagery for trail/country so it matches the project taxonomy. "Each of these is a trailhead — a marked entrance into country you already know."

Issue #977, opened 2026-05-27
Memory surfaces audit: three converging gaps in retrieval and durability


thane-developer:
Audit of how the model retrieves and writes durable knowledge, driven by a single fidelity-test failure on 2026-05-27 (the "why won't I use the game room door after a shower?" turn). The diagnosis kept widening until it surfaced three structurally distinct problems — none of them new, all of them quietly eroding the agent's ability to know its user across sessions. Filing as one issue because the fixes interact and the right ordering matters.

The investigative tool is `prewarm-replay` (#972), which now includes `compare` and `surfaces` subcommands purpose-built for this kind of A/B against production data.

## Finding 1 — `archive_search` retrieval has two correctable defects

`internal/state/memory/archive.go:2593` (`sanitizeFTS5Query`) splits multi-word queries into OR-of-quoted-tokens. For `"game room door"` it produces `"game" OR "room" OR "door"` — broader recall but loses ranking signal. Combined with BM25's preference for dense-keyword content, this puts wake-bridge `Anticipation matched: ...` events (which repeat the entity name 5-10× per message) at the top of every household-vocabulary search.

Harness data on `"game room door"`:

| Variant | anticipation | conversational | phrase-in-content |
|---|---|---|---|
| current (OR sanitizer, no filter) | 6/6 | 0/6 | 6/6 |
| OR sanitizer + drop `Anticipation matched:` | 0/6 | 6/6 | 5/6 |
| phrase-first sanitizer + drop anticipation | **0/6** | **6/6** | **6/6** |

On `"office door state"` (common individual words combining into a specific concept) the phrase-first sanitizer goes from 0/6 phrase-in-content to 6/6.

**Recommendations:**
- Add a default-on `exclude_anticipation` filter to `archive_search` (anticipation events have their own watchlist/event-tool retrieval paths; semantic search shouldn't compete with itself).
- Replace the OR-shred sanitizer with phrase-first + OR-backfill. Strictly a superset: when the phrase matches zero rows, the backfill returns the old OR shape; when it matches sparsely, OR backfill tops up to `limit`; when phrase saturates, OR is unused.

## Finding 2 — Distilled memory surfaces exist but aren't indexed

The diagnostic that landed this issue: `archive_search` never surfaces session summaries, working memory, KB summaries, or any other distilled content. Production has:

| Surface | Rows | Sample content |
|---|---|---|
[...]

PR #983, opened 2026-05-27
feat(memory): unify archive_search across raw, sessions, and working memory

thane-developer:
Refs #977 (Finding 2). **Stacks on [#982](https://github.com/nugget/thane-ai-agent/pull/982).** Merge #982 first; this PR will rebase to main automatically.

## Summary

#982 added the storage-side FTS5 indexes (`sessions_fts`, `working_memory_fts`). This PR wires them into the model's view. `archive_search` now returns a multi-surface envelope:

```json
{
  "messages":       [...],   // raw-message hits with context windows
  "sessions":       [...],   // session-summary matches (title + summary + tags)
  "working_memory": [...],   // per-conversation distilled state matches
  "truncated": false
}
```

The **same** `MemorySearcher` interface backs both the model-initiated `archive_search` tool AND the wake-time prewarm provider. They cannot drift. Any retrieval change — ranking, surface coverage, filtering — applies to both at once. (Per nugget's reminder: parallel search paths are the code smell to be on alert for; the unified `MemorySearcher` is the structural fix.)

## Composition

`MemorySearch` composes `ArchiveStore` (messages_fts + sessions_fts) and `WorkingMemoryStore` (working_memory_fts) behind a single interface:

```go
type MemorySearcher interface {
    Search(opts SearchOptions) (*SearchBundle, error)
}
```

Distilled surfaces soft-fail individually — a `sessions_fts` query error doesn't suppress the raw-message results the model would otherwise see. Only a raw-message search failure propagates.

The byte-cap fitter keeps the distilled surfaces across all truncation iterations: they're high signal per byte and cheap to keep. Only the raw-message tail shrinks under budget pressure.

## Behavior change visible to the model

Before:
```json
{"results": [/* raw messages */], "truncated": false}
```

After:
```json
{"messages": [...], "sessions": [...], "working_memory": [...], "truncated": false}
```

`talents/archive.md` is updated to teach the new shape and the "read distilled first" heuristic — `sessions[]` and `working_memory[]` are the summarizer's and metacog loop's compressed view; higher signal per byte than scrolling through raw context windows.

## Test plan

- [x] `go test -tags=sqlite_fts5 -race ./internal/state/memory/` — all pass including two new seam tests:
  - `TestMemorySearch_AllSurfacesReachModelEnvelope` — round-trips a query that has hits on all three surfaces through `FormatMultiKindResults`, verifies the envelope keeps every surface populated.
  - `TestMemorySearch_DistilledFailureDoesNotBlockMessages` — nil working memory store; raw-message results still survive.
- [x] `just ci` — full gate green (race, fmt, vet, architecture).
- [ ] Post-deploy, watch prewarm "Past Experience" blocks: when the summarizer drains the backlog (after #981) and starts populating `sessions.summary`, those hits should appear in the prewarm envelope alongside raw-message hits. Working memory matches should appear immediately on relevant queries.

## What's NOT in this PR

- Tuning the distilled vs raw weighting in `FormatMultiKindResults`. Today distilled hits always survive truncation while raw-message tails shrink — that's the right starting heuristic but may need adjustment after observation.
- A "this is what kind of result you're looking at" hint surface tag on individual hits. The envelope structure already conveys kind via the array name; per-result `kind` field would be redundant.

## Related

- #982 (depends on): adds `sessions_fts` and `working_memory_fts` + the `SearchSessions` / `WorkingMemoryStore.Search` methods this PR consumes.
- #981 (independent): unblocks the summarizer so `sessions.summary` starts getting populated again; once that lands, `sessions[]` hits in the envelope go from rare → common.

### DIFF
--- a/internal/state/memory/archive_provider.go
+++ b/internal/state/memory/archive_provider.go
@@ -10,35 +10,37 @@ import (
 	"github.com/nugget/thane-ai-agent/internal/state/knowledge"
 )
 
-// ArchiveSearcher abstracts archive search for testing. ArchiveStore
-// satisfies this interface — no adapter needed.
-type ArchiveSearcher interface {
-	Search(opts SearchOptions) ([]SearchResult, error)
-}
-
 // ArchiveContextProvider implements [agent.TagContextProvider] for
 // injecting relevant past conversation excerpts into the system
 // prompt. This is Layer 2 of the pre-warming system: Layer 1 provides
 // knowledge (facts + KB docs), Layer 2 provides experiential judgment
 // (prior reasoning about similar situations). Registered via
 // [agent.Loop.RegisterAlwaysContextProvider].
+//
+// Consumes the unified [MemorySearcher] interface so prewarm and the
+// model-initiated archive_search tool share one retrieval path — any
+// improvement to ranking or surface coverage applies to both
+// simultaneously.
 type ArchiveContextProvider struct {
-	store      ArchiveSearcher
+	searcher   MemorySearcher
 	maxResults int
 	maxBytes   int
 	logger     *slog.Logger
 }
 
-// NewArchiveContextProvider creates a context provider that searches the
-// conversation archive for relevant past exchanges. maxResults caps the
-// number of search hits; maxBytes caps the formatted output size to
-// prevent context flooding.
-func NewArchiveContextProvider(store ArchiveSearcher, maxResults, maxBytes int, logger *slog.Logger) *ArchiveContextProvider {
+// NewArchiveContextProvider creates a context provider that searches
+// every memory surface (raw messages, session summaries, working
+// memory) for content relevant to the current wake. maxResults caps
+// the raw-message hit count; the distilled surfaces use their own
+// tighter caps inside [MemorySearch.Search]. maxBytes caps the
+// formatted output size so the prewarm block never overruns the
+// system prompt's budget.
+func NewArchiveContextProvider(searcher MemorySearcher, maxResults, maxBytes int, logger *slog.Logger) *ArchiveContextProvider {
 	if logger == nil {
 		logger = slog.Default()
 	}
 	return &ArchiveContextProvider{
-		store:      store,
+		searcher:   searcher,
 		maxResults: maxResults,
 		maxBytes:   maxBytes,
 		logger:     logger,
@@ -82,7 +84,7 @@ func (p *ArchiveContextProvider) TagContext(ctx context.Context, req agentctx.Co
 	)
 
 	start := time.Now()
-	results, err := p.store.Search(SearchOptions{
+	bundle, err := p.searcher.Search(SearchOptions{
 		Query: query,
 		Limit: p.maxResults,
 	})
@@ -97,18 +99,24 @@ func (p *ArchiveContextProvider) TagContext(ctx context.Context, req agentctx.Co
 		return "", nil
 	}
 
+	totalHits := 0
+	if bundle != nil {
+		totalHits = len(bundle.Messages) + len(bundle.Sessions) + len(bundle.WorkingMemory)
+	}
 	p.logger.Debug("archive pre-warm: search completed",
 		"query", query,
-		"results", len(results),
+		"messages", len(bundle.Messages),
+		"sessions", len(bundle.Sessions),
+		"working_memory", len(bundle.WorkingMemory),
 		"took_ms", elapsed.Milliseconds(),
 	)
 
-	if len(results) == 0 {
+	if totalHits == 0 {
 		return "", nil
 	}
 
-	for i, r := range results {
-		p.logger.Debug("archive pre-warm: result",
+	for i, r := range bundle.Messages {
+		p.logger.Debug("archive pre-warm: message result",
 			"index", i,
 			"session_id", r.SessionID,
 			"match_role", r.Match.Role,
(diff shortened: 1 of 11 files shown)


internal/state/memory/archive_provider.go
```diff
@@ -97,18 +99,24 @@ func (p *ArchiveContextProvider) TagContext(ctx context.Context, req agentctx.Co
...
+		totalHits = len(bundle.Messages) + len(bundle.Sessions) + len(bundle.WorkingMemory)
+	}
 	p.logger.Debug("archive pre-warm: search completed",
 		"query", query,
-		"results", len(results),
+		"messages", len(bundle.Messages),
+		"sessions", len(bundle.Sessions),
+		"working_memory", len(bundle.WorkingMemory),
 		"took_ms", elapsed.Milliseconds(),
 	)
```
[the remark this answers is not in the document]
thane-developer: Fixed in d96d52a. Pinned the contract explicitly: MemorySearcher.Search MUST return a non-nil bundle on a nil error — "no hits" is signaled by an empty bundle, not nil. The provider now treats nil-bundle-with-nil-error as a soft programming-fault (log Warn + return empty) rather than panicking. Interface godoc updated to document the contract. MemorySearch and the test mocks already honor it.

internal/tools/archive_tools.go
```diff
@@ -105,26 +118,39 @@ func (r *Registry) registerArchiveSearch(store *memory.ArchiveStore) {
...
+				clipped := &memory.SearchBundle{
+					Messages:      bundle.Messages[:k],
+					Sessions:      bundle.Sessions,
+					WorkingMemory: bundle.WorkingMemory,
+					Truncated:     bundle.Truncated,
+				}
+				return memory.FormatMultiKindResults(clipped, now, k < len(bundle.Messages))
 			}
-			data := memory.FitPrefix(len(results), archiveResultByteCap, render)
+			data := memory.FitPrefix(len(bundle.Messages), archiveResultByteCap, render)
```
thane-developer: The byte-cap fitter only varies `len(bundle.Messages)`. If a query has zero raw-message hits but returns large session summaries or working-memory content, `FitPrefix(0, ...)` returns `render(0)` without ever checking `archiveResultByteCap`, so the tool can exceed the 16 KB cap. The prewarm formatter has an explicit distilled-only budget check; the tool path probably needs the same kind of fallback, or per-item clipping for `sessions[].summary` and `working_memory[].content` before rendering.

internal/state/memory/format.go
```diff
@@ -213,12 +234,93 @@ func FormatSearchResults(results []SearchResult, now time.Time, truncated bool)
...
+// session_id pairs with archive_session_transcript for full
+// drilldown when the match looks worth reading.
+type SessionMatchView struct {
+	SessionID      string `json:"session_id"`
+	ConversationID string `json:"conversation_id"`
+	Started        string `json:"started"`
+	Ended          string `json:"ended,omitempty"`
+	Title          string `json:"title"`
+	Summary        string `json:"summary,omitempty"`
+	Tags           string `json:"tags,omitempty"`
```
thane-developer: This stack looks stale relative to #982’s latest head. #982 fixed `SessionMatch.Tags` to be a structured `[]string` and added the `sessionsFTSEnabled` gate; this PR still introduces the model-facing view as a raw string and appears based on the pre-fix #982 commit. Please rebase #983 over the current #982 head before review/merge, otherwise the search envelope can regress the tag shape and the sessions_fts graceful-degrade fix.

internal/state/memory/distilled_fts.go
```diff
@@ -3,11 +3,118 @@ package memory
...
+
+	msgs, err := m.archive.Search(opts)
+	if err != nil {
+		return nil, err
+	}
+	bundle.Messages = msgs
+
+	// Session summaries. Soft-fail: a sessions_fts query error
+	// shouldn't drop the raw-message results we already collected.
+	if sess, err := m.archive.SearchSessions(opts.Query, maxDistilledSessions); err == nil {
```
thane-developer, line 110: This unifies the surfaces, which is the right structural move, but scoped searches now have mixed semantics: `opts.ConversationID` scopes only the raw-message search while `sessions[]` and `working_memory[]` still search globally. For model-initiated `archive_search({conversation_id: ...})`, that can inject distilled hits from unrelated conversations into a deliberately narrowed query. Both new surfaces carry `conversation_id`, so I’d either honor the scope in `SearchSessions` / `WorkingMemoryStore.Search` or suppress those surfaces when a conversation scope is present.

internal/state/memory/format.go
thane-developer: Fixed in fe53d18 (rebased on the updated #982 head). SessionMatchView.Tags is now []string, no JSON-string regression in the model-facing envelope.

internal/tools/archive_tools.go
thane-developer: Fixed in fe53d18. Replaced the single-stage FitPrefix with a two-stage shrink: stage 1 trims raw-message tail; stage 2 trims sessions then working_memory if the envelope is still over budget. The zero-message + heavy-distilled case is now handled. Defensive "force at least one hit" pass extended to all three surfaces — priority is messages, then sessions, then working_memory.

internal/state/memory/distilled_fts.go
thane-developer: Fixed in fe53d18. opts.ConversationID now scopes all three surfaces. Raw-message search filters in SQL via ArchiveStore.Search; distilled surfaces (sessions_fts and working_memory_fts are global today) get filtered in Go after the FTS pass. The filter is cheap because distilled limits are small (5 sessions, 3 working_memory) — bounded fetch-then-discard.

PR #985, opened 2026-05-27
feat(curator): bake in a memory curator loop alongside ego and metacog

thane-developer:
## What this is

A third baked-in service loop, sibling to **ego** and **metacognitive**. Where ego maintains self-reflection and metacog watches in-flight behavior, the **curator** tends thane's understanding across the memory silos by authoring **dossiers** — long-lived synthesis documents keyed by subject.

Disabled by default. Enable via:

```yaml
curator:
  enabled: true
```

## Augment, not replacement

The existing Go-side `SummarizerWorker` stays exactly as it is. It remains the clerk that stamps session metadata on session-close events. The curator is the **librarian** who turns that flow of records into per-subject dossiers. Two cadences, two roles.

| | Summarizer worker | Curator loop |
|---|---|---|
| Trigger | Session close event | Self-paced (cron-like) |
| Scope | Per-session | Per-subject |
| Output | `sessions.summary` row | Dossier as managed document |
| Cadence | Reactive (~ minutes after close) | ~1h default, self-tunable 15m–12h |
| Role | Clerk | Librarian |

## What each iteration does

1. Read prior state from `core/curator.md` — the subject queue and dossier directory.
2. Pick one subject. Either the next item in the queue, or one that feels most needed.
3. Walk the silos: `archive_search`, `recall_fact`, `contact_lookup`, the documents tools. Pull every piece of evidence for the subject.
4. Write or refresh a dossier under `dossiers/` via the documents tools, structured as claim-with-citation:

   ```markdown
   ## Claims
   - <claim> — evidence: archive:session-019c598e, fact:home/game_room_door
   - <claim> — evidence: contact:019c76e4 notes
   ```

   No claim without a citation. If you can't back it, it's an open question, not a claim.
5. Update `core/curator.md` with what was worked on and what's queued next.
6. Set the next sleep.

## Architecture

Mirrors `ego.go` exactly so the maintenance pattern is shared:

- `internal/runtime/curator/curator.go` — `DefinitionName`, `Config`, `ParseConfig`, `DefinitionSpec`, `HydrateSpec`, `BuildSpec`, `BuildLoopConfig`.
- `internal/model/prompts/curator.go` — `CuratorPrompt` (normal + supervisor augmentation).
- `internal/platform/config/config.go` — `CuratorConfig` + defaults + validation.
- `internal/platform/config/example.go` — example values for config render.
- `internal/app/app.go` + `loop_definition_hydration.go` — wiring into the app's loop-definition base specs and runtime hydration.
- Maintained output `core:curator.md` is the working state file.
- Tool exclusions match ego's exclusions: bare workspace file tools, exec, session control, tag manipulation, direct human egress. Documents tools and read tools (archive_search, recall_fact, contact_lookup) deliberately stay in.

Sleep envelope sits between metacog (minutes) and ego (hours-days). Default 1h cadence with 15m–12h range. Supervisor probability 0.1 — lower than ego's 0.2 since the curator's job is more deterministic in shape (the supervisor catches evidence-discipline drift rather than novel observation).

## What's NOT in v1

Deliberately minimal:
- **No auto-trending detector.** Subject selection is hand-driven via the queue the curator maintains in `core/curator.md`.
- **No cross-silo subject-presence index.** A future enhancement; for now the curator's evidence-gathering uses the existing search tools.
- **No metacog watchdog for fact extraction.** A future variant could scan `"noted"/"got it"` patterns and surface fact candidates; that's its own loop.
- **No formal dossier schema.** The prompt describes the recommended skeleton; the loop is free to evolve it. Iteration 1 is about seeing what shape the curator naturally produces.

These build naturally on this scaffold once production observation tells us what works.

## Test plan

- [x] `go test -race ./internal/runtime/curator/ ./internal/model/prompts/` — all pass
- [x] `just ci` — full gate green (fmt, vet, lint, race, architecture, link-check, config-generate-check). Architecture baseline bumped for the new package (57 → 58).
- [ ] **Post-merge / post-deploy on a non-prod host**: enable the curator, give it 4–6 hours, inspect:
  - Is `core/curator.md` being maintained sensibly (queue, dossier directory, notes)?
  - Are dossiers landing under `dossiers/` with claim-citation structure?
  - Does the loop converge on quiet sleeps when there's nothing to do, or does it churn?
  - Do dossier writes overlap with the existing summarizer worker output, or do they sit cleanly above it?
- [ ] **Once observation looks good**: enable on production. Watch whether interactive turns that touch curated subjects (e.g. `entity:binary_sensor.game_room_door`) start surfacing dossiers via `archive_search` (depends on #983's multi-surface envelope landing first to be model-visible).

## Strategic context

This realizes the curator-as-librarian discussion from the recent strategic round-trip. The kernel was: silos are organizational convenience for storage, but the agent's mental model is unified — make the meta-layer reflect that. The curator's dossier-keyed-by-subject output is the first concrete artifact of that unified-meta-model.

Most of the in-flight #977 work (especially #982 + #983) makes the curator's output reachable to the interactive model. Without the multi-surface FTS envelope, dossiers are just files; with it, they show up as session-summary-tier hits during semantic search. The two PR threads compose cleanly.

### DIFF
--- a/internal/model/prompts/curator.go
+++ b/internal/model/prompts/curator.go
@@ -1,0 +1,177 @@
+package prompts
+
+// curatorBaseTemplate is the prompt for each curator loop iteration.
+// Current curator.md state is injected by the loop-declared output
+// context provider as the "Declared Durable Outputs" block.
+const curatorBaseTemplate = `Curator loop iteration.
+
+You are running as a background memory curator. You tend thane's
+accumulated understanding across the memory silos — archive (past
+conversations), session summaries, working memory, facts, documents,
+contacts. The Go-side summarizer worker is the clerk that stamps
+session metadata as sessions close; you are the librarian who turns
+that flow of records into coherent dossiers keyed by subject.
+
+A dossier is a long-lived synthesis document about one subject —
+an entity (` + "`entity:binary_sensor.game_room_door`" + `), an area
+(` + "`area:kitchen`" + `), a contact (` + "`contact:<id>`" + `), a routine, a theme.
+It collects what is known about that subject across every silo and
+arranges it as **claims with citations**. The interactive agent
+reads dossiers when something jogs a memory of that subject.
+
+## Your Durable Output
+
+Your current durable output contract is injected in the "Declared
+Durable Outputs" block. That block shows core:curator.md when it
+exists and names the generated replacement tool for the document.
+
+## What To Do This Iteration
+
+1. **Read your state file** — Review your curator.md content. It
+   carries: the subject you worked on last pass, the queue of
+   subjects worth refining, and a directory of dossiers you've
+   already authored.
+2. **Pick one subject** — Either the next item in your queue, or
+   the one that feels most needed (a known fertile subject whose
+   dossier is stale, a subject with rising cross-silo presence,
+   or a new arrival that has hits everywhere but no dossier yet).
+   One subject per iteration. Depth over breadth.
+3. **Walk the silos** for that subject. Use ` + "`archive_search`" + ` for
+   raw conversation hits and session summaries; ` + "`recall_fact`" + ` for
+   stored facts; ` + "`contact_lookup`" + ` if the subject is contact-
+   shaped; the documents tools to read any existing dossier and
+   adjacent KB content.
+4. **Write or refresh the dossier** as a document under the
+   ` + "`dossiers/`" + ` namespace via the documents tools. Use the subject
+   identifier as the filename (` + "`dossiers/entity-binary_sensor-game_room_door.md`" + `
+   etc.). Structure each claim with an evidence citation — an
+   archive session ID, a fact category+key, a document ref, or a
+   working-memory conversation ID — so a reader (you, future) can
+   check any claim against its source.
+5. **Update curator.md** — Call replace_output_curator_state with
+   the complete updated body. Record what subject you worked on,
+   what came next in the queue, and any pointers to the dossier
+   you wrote/updated.
+6. **Set your sleep** — Call set_next_sleep with your chosen
+   duration. Around 1h is the default rhythm; shorter if you sense
+   an active accumulation worth following, longer if the corpus
+   feels quiet and there is nothing new to synthesize.
+
+## What a dossier should look like
+
+A short markdown document. The standard skeleton:
+
+` + "```" + `markdown
+# Dossier: <subject identifier>
+
+**Subject:** ` + "`entity:binary_sensor.game_room_door`" + `
+**Last refreshed:** <ISO date>
+**Cross-silo presence:** archive (N hits), sessions (M summaries), facts (K), working_memory (J)
+
+## Summary
+
+One paragraph synthesizing what thane currently understands about
+this subject. Write it so a fresh wake into a conversation
+mentioning the subject benefits from reading just this paragraph.
+
+## Claims
+
+- <claim> — evidence: archive:session-019c598e, fact:home/game_room_door
+- <claim> — evidence: archive:session-019c5990
+- <claim> — evidence: contact:019c76e4 notes field
+- <claim> — evidence: documents:households/layout.md
+- <claim> — evidence: working_memory:conv-signal-15124232707
+
+## Open questions
+
+- <question> — what evidence might resolve it
+- <question> — …
+
+## Connections
+
+- Related subjects: ` + "`area:game_room`" + `, ` + "`zone:smoke_break`" + `
+- Dossiers that reference this one: <links>
+` + "```" + `
+
+Every claim line carries citations. If you cannot back a claim with
+specific evidence from the corpus, do not assert it — note it as an
+open question instead. Synthesis is connecting things you can
+defend, not generating plausible-sounding text.
+
+## What you are NOT for
+
+- Writing facts on the model's behalf. The interactive agent has
+  its own ` + "`remember_fact`" + ` instinct. Your job is synthesis above
+  that layer, not replacing it.
+- Sending messages to the user or any other channel. The curator
+  is a silent process. Direct human egress tools are not available.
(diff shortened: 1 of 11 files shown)


thane-developer:
**Architecture context for reviewers:** This PR is the **synthesis layer scaffold** in the memory architecture described in #986. The two design follow-ups already filed against it:

- #989 — replace `SummarizerWorker` LLM path with curator wake events (the staged destination after this scaffold is stable)
- #988 — DRY the construction pattern across the three core service loops (ego / metacognitive / curator) + prompt-audit refresh

Both depend on this PR landing first. The umbrella issue (#986) lays out the layered story end-to-end with consistent vocabulary.

internal/model/prompts/curator.go
```diff
@@ -0,0 +1,177 @@
...
+   or a new arrival that has hits everywhere but no dossier yet).
+   One subject per iteration. Depth over breadth.
+3. **Walk the silos** for that subject. Use ` + "`archive_search`" + ` for
+   raw conversation hits and session summaries; ` + "`recall_fact`" + ` for
+   stored facts; ` + "`contact_lookup`" + ` if the subject is contact-
+   shaped; the documents tools to read any existing dossier and
+   adjacent KB content.
+4. **Write or refresh the dossier** as a document under the
+   ` + "`dossiers/`" + ` namespace via the documents tools. Use the subject
+   identifier as the filename (` + "`dossiers/entity-binary_sensor-game_room_door.md`" + `
```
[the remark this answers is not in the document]
thane-developer: Fixed in ae5a254. Updated the prompt to use `kb:dossiers/<file>.md` consistently — both in the iteration steps and in the dossier template's Connections section — and added explicit guidance that bare paths fail.

internal/platform/config/config.go
```diff
@@ -2718,6 +2804,43 @@ func (c *Config) validateEgo() error {
 	return nil
 }
 
+// validateCurator checks curator loop configuration for internal
+// consistency. No-op when the loop is disabled.
+func (c *Config) validateCurator() error {
```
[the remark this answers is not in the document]
thane-developer: Fixed in ae5a254. Added the full family of curator validation tests mirroring the ego coverage: Disabled_NoOp, RequiresWorkspace, DurationParsing (min/max/default), MinExceedsMax, DefaultOutOfRange, JitterOutOfRange, SupervisorProbabilityOutOfRange, plus the ApplyDefaults zero-floats-preserved regression guard that ego and metacog already had.

internal/model/prompts/curator.go
```diff
@@ -0,0 +1,181 @@
...
+1. **Read your state file** — Review your curator.md content. It
+   carries: the subject you worked on last pass, the queue of
+   subjects worth refining, and a directory of dossiers you've
+   already authored.
+2. **Pick one subject** — Either the next item in your queue, or
+   the one that feels most needed (a known fertile subject whose
+   dossier is stale, a subject with rising cross-silo presence,
+   or a new arrival that has hits everywhere but no dossier yet).
+   One subject per iteration. Depth over breadth.
+3. **Walk the silos** for that subject. Use ` + "`archive_search`" + ` for
```
thane-developer: One small prompt improvement for actual retrieval quality: tell the curator to search both the canonical subject handle and human aliases it discovers. A subject like `entity:binary_sensor.game_room_door` may appear in facts/config, while the archive says “game room door” or “that door with the brass handle”; phrase-first search will miss one form if the loop only queries the other. Having the curator record aliases/search terms in `curator.md` or the dossier header would also make later refreshes and the future `subjects[]` work much more reliable.

internal/model/prompts/curator.go
```diff
@@ -0,0 +1,181 @@
...
+conversations), session summaries, working memory, facts, documents,
+contacts. The Go-side summarizer worker is the clerk that stamps
+session metadata as sessions close; you are the librarian who turns
+that flow of records into coherent dossiers keyed by subject.
+
+A dossier is a long-lived synthesis document about one subject —
+an entity (` + "`entity:binary_sensor.game_room_door`" + `), an area
+(` + "`area:kitchen`" + `), a contact (` + "`contact:<id>`" + `), a routine, a theme.
+It collects what is known about that subject across every silo and
+arranges it as **claims with citations**. The interactive agent
```
thane-developer, line 19: This promise needs one more retrieval-story check. #983’s multi-surface envelope searches raw messages, session summaries, and working memory, but not KB/documents, so a `kb:dossiers/...` file written by the curator is not automatically visible through `archive_search`. Unless another current path injects/searches `kb:dossiers` on subject mention, the curator can produce good dossiers that the interactive agent rarely sees. I’d either wire/teach the model-visible retrieval path in this PR, or soften this sentence and explicitly mark dossier surfacing as depending on the document/dossier leg of the #986/#990 architecture.

internal/model/prompts/curator.go
thane-developer: Fixed in dbcc8e6. Step 3 now explicitly tells the curator to search both the canonical handle AND any human aliases (game room door / smoke-break door / brass-handle door) — phrase-first FTS misses whichever form isn't queried. Added an Aliases field to the dossier template so future passes inherit the household vocabulary from past you and don't have to re-derive it.

internal/model/prompts/curator.go
thane-developer: Filed #991 to wire dossiers (kb:dossiers/...) into the multi-surface MemorySearcher envelope from #983. You're right that the dossier-as-document is currently invisible through archive_search — the right fix is to extend MemorySearcher with a documents surface. Doing that here would balloon this PR; #991 is the focused follow-up. In the meantime, recording aliases in the dossier (the other thread) gives partial-recovery: the alias text appears in archive/working-memory hits the model already sees, even when the dossier itself isn't directly surfaced.

PR #995, opened 2026-05-27
feat: enrich native HA entity context

thane-developer:
## Summary

- Add a shared Home Assistant entity metadata projection for area, device, labels, and descriptive registry context.
- Route entity context rendering through `homeassistant/contextfmt` so `ha_get_state`, watched entities, and loop subscriptions share the same state-plus-metadata output path.
- Add `include` metadata flags to native HA entity tools and entity subscription records, including `thane_curate.entities`, `update_entity_subscriptions`, `watch_entity`, and always-visible subscriptions.
- Improve `ha_find_entity` matching with best-effort registry metadata while preserving state-only fallback when registry APIs are unavailable.
- Update HA/awareness/loop guidance for the richer native path.

## Why

Home Assistant carries rich physical-world metadata that helps the model reason about entities as things in rooms, on devices, and under labels rather than isolated state IDs. This makes the native HA path more self-sufficient and moves us closer to deprecating third-party MCP HA fallbacks.

## Validation

- `just test`
- `just ci`


internal/tools/find_entity.go
```diff
@@ -86,6 +94,22 @@ func (r *Registry) executeFindEntityHandler(ctx context.Context, argsMap map[str
 		return "", fmt.Errorf("get entities: %w", err)
 	}
 
+	lookupInclude := homeassistant.AllEntityMetadataIncludes()
```
thane-developer: **Performance: `ha_find_entity` now always fetches all four HA registries**, even when the caller passed no `include` and no `area`. `lookupInclude := AllEntityMetadataIncludes()` makes the `if lookupInclude.Any()` guard always true, so every invocation pays for entity + area + device + label registry round-trips on top of `GetEntities`.

That's a noticeable change to a hot path. The metadata-backed fuzzy match is a real win when descriptions are weak, but a lot of callers will pass a tight `description` + `domain` and never look at metadata.

Suggestion: gate the bundle fetch on `args.Include.Any() || args.Area != ""` (or a separate "needs metadata for fuzzy match" heuristic like `len(tokenize(args.Description)) >= 2`), and fall back to `fuzzyMatchEntityInfos` when no bundle is present. The downstream code already handles `metadata == nil` cleanly.

internal/tools/find_entity.go
```diff
@@ -153,30 +197,43 @@ func (r *Registry) executeFindEntityHandler(ctx context.Context, argsMap map[str
...
-		score := max(idScore, nameScore)
+		score := 0.0
+		for _, target := range targets {
+			target = strings.TrimSpace(target)
+			if target == "" {
+				continue
+			}
+			score = max(score, tokenMatchScore(descTokens, tokenize(strings.ToLower(target))))
+		}
 		if score > 0.3 { // Minimum threshold
```
thane-developer, line 228: **Fuzzy-match threshold may want re-tuning now that the target set is ~10× larger.**

Previously each entity contributed two strings (`entity_id`, `friendly_name`) to `targets`. With a metadata bundle attached, `metadataSearchTargets` adds 15+ more (description, platform, device_class, aliases, area id/name/floor/aliases, device id/name/manufacturer/model/serial/etc., every label id/name/description). The `score > 0.3` threshold and the 0.8-for-substring rule are unchanged.

That means a single-token query like `"office"` now matches any entity whose device or area or label happens to contain `office` substring with score 0.8 — much broader recall, and a higher chance of returning a confident-looking best match that's actually noise. Worth either (a) raising the threshold when a bundle is in play, or (b) weighting metadata-derived targets lower than `entity_id` / `friendly_name`, or (c) at least adding a regression test that pins down which entity wins when multiple match the description through different metadata fields.

internal/tools/ha_entity_metadata.go
```diff
@@ -0,0 +1,195 @@
...
+				"type":        "boolean",
+				"description": "Include resolved labels from the entity, device, and area.",
+			},
+			"description": map[string]any{
+				"type":        "boolean",
+				"description": "Include human-facing entity names, aliases, icon, device class, platform, category, and description when HA provides them.",
+			},
+		},
+	}
+}
```
thane-developer, line 68: **Schema/handler mismatch on the bool shortcut.**

`EntityMetadataIncludeParameter()` declares `"type": "object"`, but `ParseEntityMetadataIncludesArg` (lines 71-76) also accepts a plain `bool` — `true` is shorthand for `AllEntityMetadataIncludes()`. The schema doesn't advertise that, so:
- Strict consumers (the MCP bridge, models reading the schema literally, anything that validates args before sending) won't discover or use the bool form.
- Tool descriptions don't mention it either; they say "Use all=true for every supported field."

Two clean options:
1. **Drop the bool branch** in `ParseEntityMetadataIncludesArg` — `{"all": true}` is already the documented shorthand.
2. **Widen the schema** to `"type": ["object", "boolean"]` and add the shortcut to the description.

My preference is option 1 — one fewer shape to test and document, and `{"all": true}` is already terse.

internal/state/awareness/watchlist_store.go
```diff
@@ -443,3 +457,11 @@ func cloneTimePtr(src *time.Time) *time.Time {
...
 	return &cp
 }
+
+func cloneEntityMetadataIncludesPtr(src *homeassistant.EntityMetadataIncludes) *homeassistant.EntityMetadataIncludes {
+	if src == nil || !src.Any() {
+		return nil
+	}
+	cp := *src
+	return &cp
+}
```
thane-developer: **`cloneEntityMetadataIncludesPtr` is duplicated.**

This helper is defined identically in two packages:
- `internal/state/awareness/watchlist_store.go:461` (this file)
- `internal/app/loop_subscription_migration.go:212`

And `internal/state/awareness/loop_subscription_provider.go:148` uses the awareness-package copy. Both implementations are character-for-character the same.

Suggestion: promote to `internal/integrations/homeassistant/entity_metadata.go` as a method on the pointer receiver — e.g. `func (i *EntityMetadataIncludes) Clone() *EntityMetadataIncludes` with the same `nil`/`!Any()` short-circuit. Both consumers already import that package, and centralizing the helper makes the "copy semantics across a serialization boundary" contract explicit on the type itself.

internal/integrations/homeassistant/entity_metadata.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	if len(entry.Capabilities) > 0 {
+		meta.Capabilities = entry.Capabilities
+	}
+	if len(entry.Options) > 0 {
+		meta.Options = entry.Options
+	}
+	if len(entry.Categories) > 0 {
+		meta.Categories = entry.Categories
+	}
```
thane-developer, line 352: **Inconsistent defensive copying in `applyEntityDescription`.**

`Aliases` is defensively copied (line 205: `append([]string(nil), entry.Aliases...)`), but `Capabilities`, `Options`, and `Categories` are assigned directly from the registry entry. A consumer that mutates one of those maps on the returned `EntityMetadata` will also mutate the cached `EntityRegistryEntry`, which is shared across the whole render bundle.

This is latent today because the formatter only reads, but it's a sharp edge waiting on the first "let's tweak the metadata before encoding" use case. Either copy all collection fields for consistency, or document on the `EntityMetadata` type that it must be treated as read-only and skip the `Aliases` copy too. I'd lean toward copying — the cost is negligible for the field sizes involved and the surprise of "only Aliases is safe to mutate" is not worth the saved allocations.

thane-developer (commented):
Nice piece of work — joining area, device, and labels into one model-facing projection is a clear win for the native HA path, and the `include` shape generalizes cleanly across `ha_get_state`, `ha_find_entity`, `ha_list_entities`, watch_entity, and thane_curate.entities. The talent updates land in the right spots.

A handful of suggestions for enhancement below, in rough order of impact. None are blockers; the biggest one is the always-on registry fetch inside `ha_find_entity` (it now does 4 extra HA round-trips on every call, regardless of whether the caller asked for metadata or area filtering).

internal/tools/find_entity.go
```diff
@@ -86,6 +94,22 @@ func (r *Registry) executeFindEntityHandler(ctx context.Context, argsMap map[str
...
+	var metadata *haEntityMetadataBundle
+	if lookupInclude.Any() {
+		metadata, err = fetchHAEntityMetadataBundle(ctx, r.ha, lookupInclude)
+		if err != nil {
+			if args.Include.Any() || args.Area != "" {
+				return "", err
+			}
+			metadata = nil
+		}
+	}
```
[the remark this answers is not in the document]
thane-developer: Addressed in 4170ceda: `ha_find_entity` now derives registry lookup needs from requested `include` flags plus area filtering, and area filtering enables only area/device metadata. Added regression coverage that area lookup does not fetch the label registry.

internal/integrations/homeassistant/contextfmt/entity_format.go
```diff
@@ -18,6 +19,27 @@ func Format(state *homeassistant.State, now time.Time) string {
...
+	if metadata == nil || metadata.Empty() {
+		return formatted
+	}
+	var payload map[string]any
+	if err := json.Unmarshal([]byte(formatted), &payload); err != nil {
+		return formatted
+	}
+	payload["metadata"] = metadata
+	return promptfmt.MarshalCompact(payload)
+}
```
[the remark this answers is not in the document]
thane-developer: Addressed in 4170ceda: `AttachMetadata` now validates the compact JSON object and appends `metadata` without a map round-trip, preserving the base field order. Added an ordering regression test.

internal/tools/ha_entity_metadata.go
```diff
@@ -0,0 +1,195 @@
...
+	if b, ok := raw.(bool); ok {
+		if b {
+			return homeassistant.AllEntityMetadataIncludes(), nil
+		}
+		return homeassistant.EntityMetadataIncludes{}, nil
+	}
+	obj, ok := raw.(map[string]any)
+	if !ok {
+		return homeassistant.EntityMetadataIncludes{}, fmt.Errorf("%s must be an object with boolean metadata flags, or true for all metadata", fieldName)
+	}
```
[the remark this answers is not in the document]
thane-developer: Addressed in 4170ceda: dropped the bare bool branch so parsing matches the object-only schema. `{"all": true}` remains the documented shorthand and is covered by a regression test.

internal/state/awareness/ha_registries.go
```diff
@@ -135,6 +171,44 @@ func (r *renderRegistries) integrations() (map[string]*homeassistant.ConfigEntry
...
+	if include.Device || include.Area || include.Labels {
+		if _, err = r.devices(); err != nil {
+			return nil
+		}
+		devices = append([]homeassistant.DeviceRegistryEntry(nil), r.deviceEntries...)
+	}
+
+	resolver := homeassistant.NewEntityMetadataResolver(areas, labels, devices)
+	return resolver.MetadataForEntity(entry, state, *include)
+}
```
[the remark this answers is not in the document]
thane-developer: Addressed in 4170ceda: `renderRegistries` now caches `EntityMetadataResolver`s per include flag set for the render bundle and reuses them across entity renders. Added a call-count regression test.

internal/tools/find_entity.go
thane-developer: Addressed in 4170ceda: the bundle fetch is now gated on requested metadata or an `area` filter, with the nil-bundle fuzzy path preserved for ordinary lookups. The area path only requests area/device data and has regression coverage for no label-registry call.

internal/tools/find_entity.go
thane-developer: Addressed in 4170ceda: direct `entity_id`/`friendly_name` scoring remains primary, metadata-derived targets are weighted lower, and single-token metadata-only hits no longer clear the match threshold. Added a regression where direct `office` wins over area-only metadata.

internal/tools/ha_entity_metadata.go
thane-developer: Addressed in 4170ceda: removed the bool shortcut as suggested, leaving `{"all": true}` as the single documented all-metadata form and adding parser coverage for that shape.

internal/state/awareness/watchlist_store.go
thane-developer: Addressed in 4170ceda: promoted `(*EntityMetadataIncludes).Clone()` into the HA package and routed the watchlist store, loop provider, migration, and tool helper through that one copy path. Added nil/empty/detached-copy coverage.

internal/integrations/homeassistant/entity_metadata.go
thane-developer: Addressed in 4170ceda: `applyEntityDescription` now defensively copies `Capabilities`, `Options`, and `Categories` maps to match the existing alias-copy behavior. Added mutation-isolation coverage for all three maps.

internal/tools/ha_entity_metadata.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+	}
+
+	var devices []homeassistant.DeviceRegistryEntry
+	if include.Device || include.Area || include.Labels {
+		devices, err = ha.GetDeviceRegistry(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("get devices: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
thane-developer: Addressed/clarified in 8c866dd0 and 51cda065. We are intentionally not preserving a REST-only native HA mode here: Thane should use the HA surface that gives the model the best physical-world context. `include.area` now means richer area context, including device-inherited areas plus floor/building hierarchy, and the visibility follow-up keeps hidden-but-enabled entities available as metadata. I am keeping registry failures strict rather than silently returning a partial area-only projection.

internal/state/awareness/ha_registries.go
```diff
@@ -135,6 +179,75 @@ func (r *renderRegistries) integrations() (map[string]*homeassistant.ConfigEntry
...
+		}
+	}
+
+	var devices []homeassistant.DeviceRegistryEntry
+	if include.Device || include.Area || include.Labels {
+		if _, err := r.devices(); err != nil {
+			return homeassistant.EntityMetadataResolver{}, err
+		}
+		devices = append([]homeassistant.DeviceRegistryEntry(nil), r.deviceEntries...)
+	}
```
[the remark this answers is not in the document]
thane-developer: Addressed/clarified in 8c866dd0 and 51cda065. Render-time metadata is intentionally built from the complete HA registry hierarchy: direct area, device-inherited area, floors/buildings, labels, devices, and now visibility. A partial direct-area-only fallback would hide useful physical context from the model, so native Thane expects full HA registry access here.

internal/tools/find_entity.go
```diff
@@ -86,6 +94,23 @@ func (r *Registry) executeFindEntityHandler(ctx context.Context, argsMap map[str
 		return "", fmt.Errorf("get entities: %w", err)
 	}
 
+	lookupInclude := args.Include
+	if args.Area != "" {
+		lookupInclude.Area = true
+		lookupInclude.Device = true
+	}
```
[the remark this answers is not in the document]
thane-developer: Addressed/clarified in 8c866dd0. `ha_find_entity` keeps the device registry lookup for area filters because device-inherited area membership is part of the HA truth we want to search against. The updated coverage pins that behavior: area lookup queries the device registry, while label registry is still skipped unless labels are requested.

internal/tools/tools.go
```diff
@@ -1219,36 +1241,56 @@ func (r *Registry) handleListEntities(ctx context.Context, args map[string]any)
...
+		Domain:    domain,
+		Count:     len(matches),
+		Total:     total,
+		Truncated: total > len(matches),
+		Items:     matches,
 	}
-
-	return fmt.Sprintf("Found %d %s entities:\n%s", len(matches), domain, strings.Join(matches, "\n")), nil
+	return toIndentedJSON(result), nil
 }
```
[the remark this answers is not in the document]
thane-developer: Addressed in 8c866dd0: `ha_list_entities` now uses a list-specific truncation note suggesting a lower limit, narrower domain, or omitting include metadata rather than the automation `include_config` guidance. Added regression coverage for that envelope.

internal/integrations/homeassistant/entity_metadata.go
```diff
@@ -0,0 +1,491 @@
...
+		Aliases:             append([]string(nil), area.Aliases...),
+		FloorID:             area.FloorID,
+		Icon:                area.Icon,
+		TemperatureEntityID: area.TemperatureEntityID,
+		HumidityEntityID:    area.HumidityEntityID,
+	}
+	if floor := r.floorsByID[area.FloorID]; floor != nil {
+		out.Floor = floorMetadata(floor)
+		if r.floorAlias == "building" {
+			out.Building = floorMetadata(floor)
```
thane-developer, line 389: **Optional: the `building` alias double-emits the floor object.** When `floorAlias == "building"`, the area metadata carries both `floor` and `building` pointing at byte-identical `floorMetadata(floor)` payloads. For operators who set the alias, every area render then ships the same floor object twice, and a model reading `area.floor` and `area.building` has to wonder whether they differ (they can't). Consider emitting under a single key chosen by the alias — `building` *instead of* `floor` when aliased — so the projection stays unambiguous and doesn't pay for the duplicate. Default (`floorAlias == ""`) is unaffected since only `floor` emits there.

thane-developer (commented):
Re-audit after the revisions — this is in great shape. All five concerns from my first pass are genuinely fixed: the `Clone()` method on `EntityMetadataIncludes` replaces the duplicated `cloneEntityMetadataIncludesPtr`, `cloneAnyMap`/`cloneStringMap` close the defensive-copy gap on Capabilities/Options/Categories, `ha_find_entity` now fetches only the requested projection (no more always-on `AllEntityMetadataIncludes`), and the schema/bool-shortcut mismatch is resolved. The new visibility/salience surface is a real improvement — `visibility.context_role` as a default-context *hint* (not a hard filter) is exactly the right framing, and the per-reason area filter counts (disabled/hidden/diagnostic/config) tell the model *why* entities dropped out instead of just a total. One small optional nit below; not a blocker.

internal/integrations/homeassistant/entity_metadata.go
thane-developer: Addressed in 27cb70bd: with `floor_alias: building`, the resolver now emits the floor registry detail under `area.building` instead of duplicating the same object under both `area.floor` and `area.building`. The canonical HA area field remains `floor_id`; no synthetic `building_id` is introduced. Added regression coverage for the aliased projection.

Issue #1002, opened 2026-05-28
Native HA perception overhaul: displace high-traffic MCP HA tools with first-class native tooling


thane-developer:
## Goal

Make Thane's perception and accessibility of Home Assistant data **first-class, effortless, and enticing** — so the model reaches for native tools instead of the third-party `ha-mcp` bridge. Displace the high-traffic MCP-sourced HA tools with native implementations, then narrow the example `include_tools` allowlist to drop the now-redundant ones.

## Key finding

Mapping the native tool surface vs. the HA client vs. ha-mcp showed the **highest-value displacement wins are already data-complete** — the client already exposes `GetStates` + every registry (area/device/label/floor/category) + logbook + history. Almost nothing needs new client work. **#995 is the foundation**: its `include` metadata projection, `EntityMetadataResolver`, and visibility/salience are the substrate these tools render through.

## Sequencing

- **Phase 0 — merge #995** (the metadata substrate). Everything below composes on it; building in parallel from main would conflict heavily.
- **Phase 1 — Search** (highest traffic)
  - [ ] `ha_search_states` — predicate search across all entities (state + numeric-attribute + registry filters). The single biggest MCP-displacement win.
  - [ ] `ha_list_entities` glob + doc-drift fix — real `binary_sensor.*door*` matching; docs already claim glob the code doesn't do.
- **Phase 2 — Display / perception** (highest value)
  - [ ] `ha_area_snapshot` — "everything in the office, with states," salience-grouped (extends #995's area_activity).
  - [ ] `ha_device` — a device + all its child entities + states in one shot.
- **Phase 3 — Trends + overview**
  - [ ] Numeric history / trends — needs the one small client change (relax `no_attributes=1` on `GetStateHistory`); reuse the watchlist's summarization.
  - [ ] `ha_home_snapshot` — presence + climate + security + energy at a glance.
- **Cross-cutting**
  - [ ] Narrow the example `include_tools` MCP allowlist once Phase 1–2 land (the actual displacement).
  - [ ] Refresh `talents/ha.md` to make the native search/display paths the obvious first choice.

## Constraints

- All model-facing work follows `docs/model-facing-context.md` and `docs/model-facing-tools.md`.
- Reuse #995's `EntityMetadataIncludes` / `EntityMetadataResolver` rather than re-deriving area/device joins.
- Each tool is its own focused PR; this issue is the umbrella tracker.

## Lineage

[...]

PR #1018, opened 2026-05-28
feat(ha): ha_history — on-demand recorder trend for one entity

### LINKED ISSUE #1008: feat(ha): numeric history / trends tool (relax GetStateHistory no_attributes)
Part of #1002 (Phase 3). The one item needing a small client change.

## Problem
`GetStateHistory` forces `no_attributes=1`, so it returns state *strings* only — no numeric values. There's no native way to ask "how has the office temp trended over 24h" without the MCP history tool. The watchlist already computes min/max/avg/trend summaries on-injection; the logic exists, it just isn't exposed on demand.

## Scope
- Relax `no_attributes` on `GetStateHistory` (or add an attributes-enabled variant) — keep the lean default for the watchlist's existing callers, opt into attributes for trend work.
- New tool `ha_history` (or `ha_trend`): entity + lookback window → min/max/avg/first/last/trend + change count, reusing the watchlist summarization.
- Byte-capped; model-facing deltas.

## Acceptance
- [ ] Attributes-enabled history path on the client; existing callers unaffected.
- [ ] Tool returns numeric trend summary for numeric entities and a discrete change summary for non-numeric ones (mirroring the watchlist split).
- [ ] Tests: numeric trend, discrete history, empty window, non-recorder entity.

thane-developer:
## Summary

Adds the `ha_history` native tool — give it an entity + lookback window, get back a trend summary, without fetching raw history yourself:

- **Numeric** entities → min/max/start/end/delta + rising/falling/flat.
- **Non-numeric** entities → change count + recent states.
- Defaults to a 24h window, clamped to 30 days. Byte-lean, model-facing deltas.

It reuses the *exact* numeric/discrete summarization the always-on watchlist injection already computes — `buildWatchlistHistorySummaries` is refactored to share its post-fetch half (`summarizeHistorySeries`) rather than growing a parallel copy.

### Client change (the issue's core ask)
`GetStateHistory` keeps its lean `no_attributes` default, so **existing watchlist callers are unaffected**. A new `GetStateHistoryWithAttributes` sibling fetches the attribute map. The tool uses it **only when an `attribute` is requested**.

### A note on the issue's framing
The issue says `no_attributes=1` means "state strings only — no numeric values." That's not quite right: `no_attributes` strips only the *attribute map*, not the state value, so for a numeric sensor the state string *is* the number — the watchlist already trends those fine. The real gap the attributes path unlocks is trending a numeric value that lives in an **attribute** (e.g. a `climate` entity's `current_temperature` while its state is `"heat"`). So `ha_history` grows an optional `attribute` param for exactly that, which is what makes the new client method genuinely load-bearing rather than fetching bytes the summarizer ignores.

Closes #1008. Part of #1002 (Phase 3).

> Stacked on #1016 (`claude/1007-ha-device`). Auto-retargets to `main` as the stack lands.

## Test plan
- [x] `just ci` passes locally (architecture baseline: interfaces 77→78 for `TrendHistoryClient`).
- [x] Numeric trend (rising, min/max, unit) — `entity_trend_test.go`.
- [x] Discrete history (change count + recent states).
- [x] Empty window → explicit `no_history` marker (not a bogus 1-sample flat trend).
- [x] Non-recorder entity → `no_history`.
- [x] Attribute trend uses the attributes-enabled path and summarizes the attribute series.
- [x] Lookback clamped to 30 days before the recorder query.
- [x] Client: `GetStateHistoryWithAttributes` omits `no_attributes` and preserves attributes; `GetStateHistory` path byte-identical (existing test unchanged).

### DIFF
--- a/internal/state/awareness/entity_trend.go
+++ b/internal/state/awareness/entity_trend.go
@@ -153,20 +153,25 @@ func projectAttributeSeries(states []homeassistant.State, attribute string) []ho
 
 // projectAttributeState returns a copy of current with its state value
 // replaced by the named attribute's value, preserving the attribute map
-// so precision/unit resolution still works. Returns current unchanged
-// when the attribute is absent or non-coercible (the empty/garbage
-// series then yields a clean no-history marker).
+// so precision/unit resolution still works. Returns nil when the
+// attribute is absent or non-coercible on the current state: the shared
+// summarizer appends the current sample to the series, so handing back
+// the *unprojected* state (e.g. a climate entity's "heat") would
+// contaminate a projected numeric attribute series and force a discrete
+// fallback. Dropping the current sample instead lets the recorded
+// attribute history summarize cleanly (or yield a no-history marker
+// when the recorder also has nothing for the attribute).
 func projectAttributeState(current *homeassistant.State, attribute string) *homeassistant.State {
 	if current == nil {
 		return nil
 	}
 	raw, ok := current.Attributes[attribute]
 	if !ok {
-		return current
+		return nil
 	}
 	value, ok := attributeValueString(raw)
 	if !ok {
-		return current
+		return nil
 	}
 	next := *current
 	next.State = value
--- a/internal/state/awareness/entity_trend.go
+++ b/internal/state/awareness/entity_trend.go
@@ -1,0 +1,194 @@
+package awareness
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/nugget/thane-ai-agent/internal/integrations/homeassistant"
+	"github.com/nugget/thane-ai-agent/internal/model/promptfmt"
+)
+
+const (
+	defaultTrendLookback = 86400   // 24h
+	maxTrendLookback     = 2592000 // 30d — bounds recorder load; clamped, not rejected
+)
+
+// TrendHistoryClient is the slice of Home Assistant calls the ha_history
+// tool needs: the current state (for classification, precision, and
+// unit) plus recorder-backed history. The attributes-enabled history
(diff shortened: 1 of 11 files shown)


internal/state/awareness/entity_trend.go
```diff
@@ -0,0 +1,194 @@
...
+// replaced by the named attribute's value, preserving the attribute map
+// so precision/unit resolution still works. Returns current unchanged
+// when the attribute is absent or non-coercible (the empty/garbage
+// series then yields a clean no-history marker).
+func projectAttributeState(current *homeassistant.State, attribute string) *homeassistant.State {
+	if current == nil {
+		return nil
+	}
+	raw, ok := current.Attributes[attribute]
+	if !ok {
```
thane-developer, line 169: [P2] Returning `current` unchanged here can poison attribute trends when the recorder has numeric samples for the requested attribute but the current state no longer exposes that attribute. `summarizeHistorySeries` will append the unprojected current state (for example `heat`) to the projected numeric series, causing `summarizeNumericHistory` to fall back and the tool to return a discrete summary over numeric strings plus the entity state instead of a numeric trend or a clean no-history marker. Consider omitting the current sample for attribute trends when the current attribute is absent/non-coercible, and add a regression where history has the attribute but current does not.

thane-developer: Good catch — fixed in 4f28810. `projectAttributeState` now returns nil (drops the current sample) when the attribute is absent/non-coercible on the current state, so the unprojected entity state can't contaminate the projected numeric series. Added `TestComputeEntityTrend_AttributeAbsentOnCurrent`: recorder carries current_temperature but the current state is off → asserts kind=numeric (not a poisoned discrete) with max_value from the recorded series.

PR #1019, opened 2026-05-28
feat(ha): ha_home_snapshot — curated whole-home overview

### LINKED ISSUE #1006: feat(ha): ha_home_snapshot — whole-home overview at a glance
Part of #1002 (Phase 3). Lowest urgency; composes on the rest.

## Why
A single "how's the house right now" view — presence, climate, security/openings, energy — with no native equivalent. Composable from existing client data + the Phase 1/2 primitives.

## Tool: `ha_home_snapshot`
- Curated cross-domain overview: who's home (presence), climate per area, open doors/windows + locks, anything unavailable/anomalous, optional energy.
- Built on `ha_search_states` / `ha_area_snapshot` primitives rather than re-deriving.
- Salience-first: lead with what's actionable or anomalous.

## Acceptance
- [ ] `ha_home_snapshot` registered under `ha`.
- [ ] Sections: presence, climate, openings/security, anomalies/unavailable (energy optional).
- [ ] Reuses Phase 1/2 primitives; no duplicated joins.
- [ ] Tests for section assembly + the all-quiet path.

thane-developer:
## Summary

Adds `ha_home_snapshot` — a single curated "how's the house right now" view across domains, the capstone of the #1002 native-HA fan. Leads with what's actionable:

- **anomalies** — offline / in alarm (smoke, CO, gas, jammed lock, …)
- **security** — open doors/windows/garage, unlocked locks, armed/triggered alarm panels
- **presence** — who's home vs away (`person` entities)
- **climate** — thermostats
- **energy** — optional (`include_energy`), power/energy sensors

A top-level `summary` carries the counts at a glance, and `status: "quiet"` means nothing is offline, open, unlocked, or armed. Each section is capped (`<section>_truncated_count`). Optional `include` adds per-entity metadata.

### Built as the area snapshot's cousin (per request)
`ComputeHomeSnapshot` deliberately tracks `ComputeAreaActivity` — the implementation #1015 enhanced — rather than inventing its own shape: one bulk `GetStates`, the same registry resolver, the same salience predicates (`isSentinelState`/`isAlarmAnomaly`), the same `renderEntityForArea` + `attachAreaEntityMetadata`, the same bounded output with `*_truncated_count`. The shared salience filter (disabled/hidden/diagnostic/config) is **extracted** from `selectAreaMembers` into `keepAfterSalienceFilter` so both snapshots apply identical visibility rules — no duplicated joins. Unlike the area view it's **curated**: entities that fit no section are skipped, not bucketed into a remainder, so it stays glanceable.

Closes #1006. Completes #1002 (Phase 3).

> Stacked on #1018 (`claude/1008-ha-history`). Auto-retargets to `main` as the stack lands.

## Test plan
- [x] `just ci` passes locally (no architecture bump — reuses `HARegistryClient`, no new interface).
- [x] Section assembly: each domain routes to the right section; a plain light is curated out.
- [x] All-quiet path: `status: "quiet"`, no anomalies/security sections, presence/climate still render.
- [x] Energy section gated behind `include_energy`.
- [x] Diagnostic/disabled entities filtered + counted (shared filter).
- [x] Section cap → `<section>_truncated_count`.
- [x] `include` metadata attaches to section entities.
- [x] `selectAreaMembers` refactor leaves all existing area-snapshot tests green.

### DIFF
--- a/internal/state/awareness/home_snapshot.go
+++ b/internal/state/awareness/home_snapshot.go
@@ -1,0 +1,319 @@
+package awareness
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/nugget/thane-ai-agent/internal/integrations/homeassistant"
+	"github.com/nugget/thane-ai-agent/internal/model/promptfmt"
+)
+
+// homeSnapshotDefaultMaxPerSection caps each section so a large install
+// can't flood the prompt. Overflow is reported per section via
+// <section>_truncated_count, mirroring the area snapshot's bucket caps.
+const homeSnapshotDefaultMaxPerSection = 12
+
+// homeOpeningClasses are the binary_sensor device_classes whose "on"
+// state means something is physically open — the security/openings view.
+var homeOpeningClasses = map[string]bool{
+	"door":        true,
+	"window":      true,
+	"garage_door": true,
+	"opening":     true,
+}
+
+// homeEnergyClasses are the sensor device_classes surfaced in the
+// optional energy section.
+var homeEnergyClasses = map[string]bool{
+	"power":  true,
+	"energy": true,
+}
+
+// HomeSnapshotRequest describes one ha_home_snapshot invocation.
+type HomeSnapshotRequest struct {
+	IncludeDiagnostic bool                                 // include diagnostic/config entities in the scan
+	IncludeHidden     bool                                 // include hidden-but-enabled entities
+	IncludeEnergy     bool                                 // add the optional energy section
+	MaxPerSection     int                                  // per-section cap; <= 0 uses the default
+	Include           homeassistant.EntityMetadataIncludes // optional per-entity metadata projection
+}
+
+// ComputeHomeSnapshot renders a single curated "how's the house right
+// now" view across domains: anomalies first (offline / alarm), then
+// security/openings (open doors-windows, unlocked locks, armed/triggered
+// alarm panels), then presence (who's home), climate (thermostats), and
+// an optional energy section. It is the whole-home cousin of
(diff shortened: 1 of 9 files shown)


internal/state/awareness/home_snapshot.go
```diff
@@ -0,0 +1,319 @@
...
+
+	allStates, err := client.GetStates(ctx)
+	if err != nil {
+		return "", fmt.Errorf("ha_home_snapshot: get states: %w", err)
+	}
+	entities, err := client.GetEntityRegistry(ctx)
+	if err != nil {
+		return "", fmt.Errorf("ha_home_snapshot: get entity registry: %w", err)
+	}
+	devices, err := client.GetDeviceRegistry(ctx)
```
thane-developer: [P2] The default `ha_home_snapshot` path now always pulls the full device registry, and `buildHomeResolver` then fetches areas, even when `include` is empty. In that no-include path the snapshot only needs live states plus the entity registry: `deviceByID` is only stored on `areaMember`, classification reads the entity registry/live state, and metadata attachment is skipped. On a large HA install this adds avoidable registry round-trips to the common glanceable overview. Consider gating the device registry/resolver work on `req.Include.Any()` (or the specific include flags) so the default snapshot stays lean.

thane-developer: Good catch — fixed in b191154. The device registry fetch and the resolver build are now gated on `req.Include.Any()`, so the default glanceable snapshot does just GetStates + GetEntityRegistry (`selectHomeMembers` tolerates a nil deviceByID, and areaMember.device is unused in this view). Added tests asserting zero device/floor registry calls on the no-include path, and that the device registry IS fetched when an include projection is requested.

Issue #939, opened 2026-05-26
FTS5 index over interactions/ with archive reindex command


thane-developer:
## Background

Foundation: #937 (CLOSED — archive directory skeleton landed via PR #952). Normalizer: #938 (OPEN — must populate the `interactions/` corpus before this index has anything to derive from). Once `interactions/` records are landing on disk, we need a search index to make them fast to query. This issue covers `meta/fts/index.db` — the FTS5 sqlite over the interactions corpus, plus the `thane archive reindex` command that rebuilds it from scratch.

The index is fully derived: nuking and rebuilding it from `interactions/` must yield an identical search surface. This is the rebuildable-index invariant in code.

> **Status (2026-06-25): not started.** No `meta/fts/index.db`, no `thane archive reindex` command, no live indexer. Blocked on #938: the interactions corpus is still unpopulated (`archive/meta/schema/interactions.v1.json` is a placeholder; message fidelity currently stays in the thane.db sqlite content path). Today's FTS5 (`messages_fts`/`sessions_fts` in `internal/state/memory/`) indexes the legacy content tables — the pre-bridging surface this index supersedes.

## In scope

1. **FTS5 schema** over interactions records:
   - Indexed columns: `text` (primary FTS field), plus filterable: `kind`, `source`, `session_id`, `conversation_id`, `speaker`, `trigger`, `ts`
   - Min/max stats on `ts` for time-range pruning
2. **Incremental indexing.** As new interactions records land, they get indexed live. Drain via a buffered channel pattern, similar to existing `indexer.go`.
3. **`thane archive reindex` CLI command.** Walks `interactions/`, rebuilds `meta/fts/index.db` from scratch. Atomic: write to `.tmp`, fsync, rename. Reports row counts and verifies the rebuild matches the source corpus.
4. **Periodic WAL checkpoint** to keep the WAL file bounded.

## Acceptance criteria

- [ ] `meta/fts/index.db` schema published in `meta/schema/`
- [ ] Live indexing keeps up with current write rate without blocking writers
- [ ] `thane archive reindex` rebuilds the index from `interactions/` deterministically
- [ ] Re-running reindex on a freshly-rebuilt index is a no-op (or trivially fast verify)
- [ ] WAL checkpoint runs on a schedule
- [ ] Tests: live ingest, full rebuild, query correctness across `kind`/`source`/time filters

## Implementation constraints (from PR #1077, modernc SQLite migration)

[...]

PR #1020, opened 2026-05-28
feat(memory): archive_search time-range facet + surfaced relevance

### LINKED ISSUE #943: archive_search: time-range facet + surfaced relevance (re-scoped)
## Background

**Re-scoped 2026-05-28** against current `main`. The original premise — `sanitizeFTS5Query()` ORing all terms into a noisy, unrankable result set — has been largely overtaken by work that landed since (#977 Finding 1, #982, #983, plus the anticipation-contamination fix):

- `sanitizeFTS5Query` is **gone**. `ArchiveStore.searchFTS` now runs a **phrase-first + OR-of-terms backfill** with BM25 `ORDER BY rank` and an anticipation-contamination filter — phrase precision at the top, recall headroom only when the phrase underfills.
- `archive_search` is now **multi-surface**: `messages[]` (raw hits + gap-bounded context), `sessions[]` (distilled per-session summaries), `working_memory[]` (per-conversation distillation), all via the unified `MemorySearcher` the wake-time prewarm provider also consumes — so the two retrieval paths can't drift.
- Foundation #937 (layout) is closed; #939 (FTS5 index + reindex over `interactions/`) remains open — but is **not** a blocker (see Dependencies).

The precision/phrase goals are effectively met by a different (and better) mechanism. What remains is a smaller, facet- and signal-shaped slice.

## In scope (trimmed)

1. **`time_range` facet on `archive_search`** — `min_time` / `max_time` accepting RFC3339 or signed deltas (mirroring `archive_range`), so a *semantic* search can be scoped to a window without falling back to `archive_range`, which has no text search. Applies to the raw-message surface; distilled surfaces stay unscoped, consistent with how `conversation_id` already behaves.
2. **Surface relevance** — thread the BM25 rank through `scanFTSMatches` → `SearchResult` into the envelope as a `score`, and document that result order already reflects relevance. Add a cheap `total_estimated` hit count so the model can gauge overflow without a cursor.
3. **Description + eval refresh** — update the tool description for the new facet/score; add an eval case for time-scoped semantic search.

## Done / dropped (from the original scope)

- ✅ **Phrase support** — phrase-first is now the default pass; tested by `TestSearch_PhraseFirstPreferred`. (User-supplied quotes are escaped rather than treated as sub-phrases — the whole query is phrase-anchored. Description nuance, not new code.)
- ❌ **AND-default semantics** — dropped. Pure AND would remove the recall headroom phrase-first deliberately keeps; precision is already achieved via phrase-anchoring + BM25 ranking.
- ❌ **Cursor / pagination** — deferred. A cursor across three surfaces plus the two-stage byte-cap fitter is far more complex than originally framed, and the high-signal distilled surfaces blunt the need. The narrow-and-retry + `truncated=true` model stays; `total_estimated` (above) is the salvaged piece.
- ↪️ **Other facets** (`participant`, `kind`, `source`, `session_id`) — relocated/covered elsewhere: `conversation_id` already on `archive_search`; `time_range` + conversation on `archive_range`; tag/title browse on `archive_sessions`; `session_id` via `archive_session_transcript`. `participant` / `source` don't map cleanly to the current schema — out of scope unless a concrete need surfaces.

## Acceptance criteria

- [ ] `archive_search` accepts `min_time` / `max_time` (RFC3339 or signed delta); raw-message hits scoped to the window, distilled surfaces unscoped
- [ ] Envelope surfaces a per-hit `score` (BM25 rank) and a `total_estimated` count
- [ ] Tool description updated; existing eval cases continue to pass
- [ ] New eval case covers time-scoped semantic search

## Dependencies

**None blocking.** This ships against the current `messages_fts` backend today. When #939 (FTS5 index over `interactions/`) lands, the query internals re-target the new backend *behind the stable `MemorySearcher` interface* — the tool schema, envelope, `score`, and `time_range` facet port forward unchanged (per #939's own note: "the unified read surface ports forward cleanly; pre-bridging work is not wasted"). The `time_range` semantics should match #939's eventual `ts` column when it arrives.

## Out of scope

- Watchlist primitive (separate issue, deferred)
- Summarizer changes (deferred)
- AND-default semantics and cursor/pagination (see Done / dropped)

thane-developer:
## Summary

Implements the re-scoped #943 — the slice that's still real after the phrase-first `searchFTS` and multi-surface envelope landed (see the issue's re-scope note). Ships against the current `messages_fts` backend; the tool schema/envelope port forward behind `MemorySearcher` when #939 lands.

- **`time_range` facet** — `min_time`/`max_time` on `archive_search` (RFC3339 or signed delta like `-7d`, mirroring `archive_range`) scope the **raw-message** surface to a window. Distilled surfaces stay unscoped, exactly like `conversation_id`. Applied on **both** the FTS and LIKE paths so the facet never silently no-ops.
- **Surfaced relevance** — each message hit carries a `score` (negated BM25, higher = better) and a `match_type` (`phrase` precision pass vs `terms` OR-of-terms backfill). Scores are comparable *within* a `match_type`, not across — `match_type` makes that explicit. Results were already ordered best-first.
- **`total_estimated`** — `SearchBundle.TotalMessages` via `ArchiveStore.CountMatches` (broadest OR-of-terms recall count), so the model can gauge overflow without a cursor.

### Notable implementation detail
The same `Search`/`runFTSQuery` code serves **both** storage modes — legacy (`archive_messages`, RFC3339Nano "T"-separated) and unified (`messages`, go-sqlite3 native space-separated). A raw string compare on timestamps would silently mismatch at window edges in one mode (`" " 0x20 < "T" 0x54`). The time bounds compare through SQLite `datetime()` on **both sides**, making the filter storage-format agnostic. A shared `ftsConditions` builds the WHERE for both `runFTSQuery` and the count, so the candidate set and its total can't diverge.

Closes #943.

## Test plan
- [x] `just ci` passes locally (runs the LIKE fallback — no `sqlite_fts5` tag).
- [x] `time_range` scoping verified on the LIKE path (runs in CI) **and** under `-tags sqlite_fts5`.
- [x] `score` > 0 and `match_type == "phrase"` verified under `-tags sqlite_fts5`; the test skips on the LIKE fallback (BM25/phrase are FTS-only).
- [x] `total_estimated` / `CountMatches` verified under `-tags sqlite_fts5`; skips on LIKE.
- [x] Existing memory + archive-tool tests stay green (envelope/scan changes).


internal/tools/archive_tools.go
```diff
@@ -102,6 +107,15 @@ func (r *Registry) registerArchiveSearch(searcher memory.MemorySearcher) {
 					"type":        "string",
 					"description": "Optional: scope the raw-message search to one conversation. Distilled surfaces are unscoped. Omit to search across everything.",
 				},
+				"min_time": map[string]any{
+					"type": "string",
+					"description": "Optional: only return raw-message hits at or after this time. Accepts RFC3339 " +
+						"(\"2026-04-25T14:00:00Z\") or a signed delta (\"-7d\", \"-3600s\"). Scopes the raw-message surface only; distilled surfaces stay unscoped.",
```
thane-developer, line 113: [P2] Don’t advertise `-7d` unless the parser accepts it. These args flow through `promptfmt.ParseTimeOrDelta`, which currently accepts RFC3339 or signed seconds (`+3600s`/`-300s`) only. If the model follows this example and sends `-7d`, the tool returns a parse error instead of the intended week-scoped semantic search. Either extend the parser/coverage for day-style deltas or change these examples to the signed-second form, e.g. `-604800s`.

internal/state/memory/format.go
```diff
@@ -232,6 +240,8 @@ func buildSearchResultViews(results []SearchResult, now time.Time) []SearchResul
 			ContextBefore: searchContextViews(tailMessages(r.ContextBefore, maxSearchContextPerSide), now),
 			ContextAfter:  searchContextViews(headMessages(r.ContextAfter, maxSearchContextPerSide), now),
 			Highlight:     r.Highlight,
+			Score:         math.Round(r.Score*100) / 100,
```
thane-developer: [P2] This rounding will usually erase the relevance signal before it reaches the model. SQLite FTS5 `bm25()` scores are commonly very small (for example, a one-row exact phrase returns about `-1e-06`; after negation that becomes `0.000001`), so `math.Round(score*100)/100` renders most real hits as `0`. The raw `SearchResult.Score > 0` test still passes, but the JSON field this PR adds becomes uninformative. Please preserve more precision or scale/normalize before rounding, ideally with a formatter-level test that catches non-zero BM25 scores surviving into the envelope.

internal/tools/archive_tools.go
thane-developer: Fixed in f03bd2a. Rather than downgrade the example, I extended `ParseTimeOrDelta` to accept `<sign><int><unit>` with unit s/m/h/d/w (backward compatible — `-300s` still parses), so `-7d` is now valid as advertised. This also lets archive_range and any other ParseTimeOrDelta caller take human-scale windows. Added parser coverage for the new units (-30m/-24h/-7d/+2w) plus an unknown-unit error case.

internal/state/memory/format.go
thane-developer: Good catch — fixed in f03bd2a. Switched from fixed 2-decimal rounding to 4 *significant figures* (`roundScoreSignificant`), so the leading digits survive at any magnitude — a ~1e-06 one-row phrase score stays 0.000001 in the envelope instead of flattening to 0. Added a formatter-level test (`TestFormatMultiKindResults_PreservesSmallScore`) that feeds a 1e-06 score through FormatMultiKindResults and asserts the JSON score is non-zero (and that match_type/total_estimated reach the envelope). It's a pure formatter test, so it runs in CI without the sqlite_fts5 tag.

Issue #1024, opened 2026-05-28
Loop fork bomb from unified spawning — move background services (curator, MQTT, …) onto a shared queue-consumer primitive


thane-developer:
> 🧨 The fork-bomb issue landed on **#1024 = 2¹⁰**. We did not plan this. The universe has a sense of humor about self-replication.

## TL;DR

A recent relaxation — unifying loop toolsets so any turn can `thane_*` new loops (delegates were previously barred from launching delegates) — turned the **curator** into a fork bomb. It self-replicated (`loop-curator-1…17`), woken every ~5 min, firing a router-bypassing stream of LLM calls onto the one slow local model (`gpt-oss:120b`/spark) that interactive turns also depend on, and **DoSed interactive Signal turns** (timeouts at the ~10-min deadline).

The fix is **not** to re-add the old binary "delegates can't spawn" rule. It's to recognize that we now have **two services of the same shape (MQTT wakes, curator) and almost certainly more coming**, and to build **one primitive** that contains them: a **durable queue drained by a pre-defined consumer loop**, organized under a **container node linked to core**. Spawning stays available for genuine interactive orchestration, but under a bounded budget.

## Root cause: a fork bomb, not just a busy loop

The unification (#902) collapsed two *different* workload classes onto a single mechanism — spawning:

1. **Event-driven / recurring background work** (curator, summarizer, MQTT wakes, ego, metacognitive) — a *stream* of "process this" jobs.
2. **Interactive hierarchical orchestration** (an owner turn delegating sub-work) — a *tree* of delegations.

Spawn-per-event is the textbook wrong tool for class (1): no backpressure, unbounded concurrency, self-replication. The old "delegates can't spawn delegates" rule was a blunt guard on the wrong axis (identity — *are* you a delegate?) instead of the real axis (a finite spawn budget). The relaxation set that budget to ∞ for everyone, including background loops that should never have been spawning at all.

## Direction: one shared queue-consumer primitive

MQTT-wake migration already moved class (1) work from *spawn-per-event* to *enqueue into a pre-defined loop*. Curator is the second, structurally identical instance. **This is the cue to build the primitive once** rather than per service:

> **Trigger → enqueue (deduped/coalesced, admission-controlled) → durable queue → one pre-defined consumer loop draining at a controlled rate on a non-interactive model lane.**

[...]

Issue #989, opened 2026-05-27
Replace SummarizerWorker LLM path with curator wake events


thane-developer:
## Background

Architecture context: #986 (umbrella).

The current `SummarizerWorker` is a Go-side background worker that polls for unsummarized sessions and, for each, routes a model + calls LLM + parses metadata + writes `sessions.summary`/`title`/`tags`. It works mechanically, but it produces a **second model voice** — separate from the curator's voice — handling session-level synthesis.

Two parallel synthesis paths is the code smell #977's findings kept surfacing. The destination is **one librarian, two stimuli**: the curator handles both its self-paced wake (pulse) and session-close events (pulse-driven).

## Destination flow

Each session close fires a wake event into the curator loop with payload `{session_id, kind: session_close}`. The curator:

1. Reads the session transcript (already-archived).
2. Decides what changed.
3. Folds new evidence into relevant dossiers.
4. Writes session metadata (`sessions.summary`/`title`/`tags`) as a side effect — or skips with `markEmpty` when the session is trivial.
5. Sleeps until the next event or its own self-paced wake.

Same model voice, same prompt foundation, two stimuli.

## Dependencies

- **#985** (curator loop scaffold) — must be merged and stable.
- **#938** (interactions normalizer) — ideally lands first. At that point the natural emit site for "session closed" is the interactions normalizer when it writes a `kind: session` record, not the existing `SummarizerWorker`.

### Two implementation paths

**If #938 lands first:** the interactions normalizer emits the curator wake when it writes a `kind: session` record. `SummarizerWorker` deletes outright.

**If this lands before #938:** `SummarizerWorker` becomes a thin event-firing shim — listens for session-close, fires the wake, no LLM call. Retires fully when #938 ships.

## In scope

1. **Curator prompt grows a two-modes section.** Self-paced and session-close-event. Each gets a few sentences of guidance distinguishing the work.
2. **Wake-event plumbing** for session-close → curator. Reuse existing event-driven loop machinery from `internal/runtime/loop/`.
3. **`SummarizerWorker` transformation:** either delete (if #938 emit site exists) or convert to thin event emitter (no LLM call).
4. **Tests** covering the event delivery path + curator's mode switch on payload.
5. **Backpressure / flood handling.** Session-close storms (mass-archive runs, crash recovery sweeps) must not flood the curator. Batching policy needed.

## Acceptance

- [ ] No LLM call from `SummarizerWorker`.
[...]

PR #1025, opened 2026-05-29
feat(archivist): self-paced queue consumer replaces event-woken curator (#1024)

thane-developer:
## Summary

Fixes the production incident in #1024: every session close notified the **curator** loop (`NotifyLoopByName`), re-triggering a full dossier-crawl iteration every few minutes. On `gpt-oss:120b`/spark — the only routable local model, which interactive turns also use — that pinned the GPU and timed out Signal turns. The `loop-curator-N` IDs were **iterations of one loop**, so this was **event-wake amplification**, not recursive spawning.

The fix generalizes the in-memory loop-notification mechanism into a **durable, deduped, per-consumer-loop work queue** — the pull-based form of a notification. Producers `Enqueue` work for a named consumer loop; the consumer drains its own partition on its own cadence. **Trigger rate no longer drives work rate.** The curator is renamed **archivist** (it shared a blurry border with the `thane_curate` tool) and is the queue's first consumer; MQTT is the designed next one.

## What's in it

- **`internal/state/loopqueue`** — generic queue store, partitioned by `consumer_loop`, deduped on `(consumer_loop, dedup_key)`. Payload is a `messages.LoopNotifyPayload` (no new types). `Enqueue` upserts/coalesces (priority MAX, `enqueued_at` preserved → no starvation); `Peek` is read-only (single consumer + ack-on-done = at-least-once, crash-safe); `Ack` deletes.
- **`internal/app/loop_queue_tools.go`** — `queue_pull` / `queue_ack` / `queue_enqueue` as **loop-private `RuntimeTools`** closed over the consumer's own name (its own partition only). Never registered globally — nothing else can drain or enqueue into a loop's partition.
- **curator → archivist rename** — package, config key (`curator:` → `archivist:`), state file (`curator.md` → `archivist.md`), prompt, mission. Prompt rewritten to a single **drain-your-queue** mode (no event-wake branches; removed the false "you write session metadata" claim).
- **Zero spawn rights** for the archivist (excludes `spawn_loop`/`thane_now`/`thane_assign`/`thane_curate`/`thane_create_container`) — the per-class half of the fork-bomb guard. Frontier expansion is `queue_enqueue`, not a spawn.
- **Producer rewire** — session close enqueues into the archivist partition instead of notifying; the loop is no longer event-woken.
- **Summarizer decoupling** — dropped the #989 short-circuit. The summarizer now **always** stamps session metadata AND enqueues for the archivist as a side effect, fixing the latent bug where a successful wake left `title` NULL and the session re-scanned forever (a second amplifier).

## Designed-but-deferred seam

The `WakeOnEnqueue` policy (event-driven consumers like MQTT: an enqueue fires a *debounced* wake that drains a batch) is designed in — `Enqueue` is the single wake chokepoint — but only the self-paced path ships here. The debounced-wake path + MQTT migration land as the follow-up that first needs it (see #1024 v1-scope comment).

## ⚠️ Breaking config change

The config key is renamed **`curator:` → `archivist:`**. Update prod `config.yaml` in lockstep when promoting; validate with `thane -config <candidate> validate` before restart. The old `core/curator.md` state doc is orphaned (a new `archivist.md` is created) — harmless.

## Test plan

- [x] `just ci` green locally (fmt, vet, config-generate-check, architecture-check — baseline bumped for the new `loopqueue` package, build, race tests)
- [x] loopqueue store: enqueue/coalesce (priority MAX, partition isolation), Peek ordering, Ack delete
- [x] queue tools: pull/ack/enqueue handlers, partition scoping
- [x] `enqueueSessionCloseWork`: closed session → `session:<id>` in the archivist partition, coalesces on re-enqueue, errors on empty id
- [x] summarizer: always summarizes + enqueues as a side effect; enqueue failure doesn't block metadata
- [x] archivist spec: excludes spawn tools; prompt is single drain-queue mode
- [ ] Prod after deploy: close a session → a `loop_queue` row appears (`consumer_loop=archivist`), **no** `loop-archivist-N` iteration is triggered by the close, archivist drains on its timer, session title is written

Refs #1024


internal/state/loopqueue/queue_store.go
```diff
@@ -0,0 +1,162 @@
...
+//
+// dedupKey is the per-consumer identity of the work (e.g.
+// "session:<id>", "entity:<id>"): a second Enqueue with the same
+// (consumerLoop, dedupKey) refreshes the payload and re-arms the item
+// rather than appending a duplicate. The original enqueued_at is
+// preserved so a repeatedly-touched subject keeps its FIFO position and
+// can't starve older work; priority is raised to MAX(old, new) so a
+// high-priority re-enqueue can promote but a low-priority duplicate
+// can't silently demote.
+func (s *Store) Enqueue(consumerLoop, dedupKey string, priority int, payload []byte) error {
```
thane-developer: [P1] The queue primitive is missing the v1 backpressure/admission guard. Enqueue dedupes identical keys, but a frontier expansion that keeps discovering new keys can still grow loop_queue without bound; queue_enqueue gives the archivist exactly that ability. #1024 calls out max_depth+overflow as one of the load-bearing v1 attributes, because otherwise the failure mode just moves from live loop storm to durable queue depth. Please enforce a per-consumer depth cap transactionally, only for new keys; coalesces should still update, and return/surface an overflow result so the consumer knows work was shed.

internal/state/loopqueue/queue_store.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return nil
+}
+
+// Peek returns up to limit pending items for consumerLoop in drain order
+// (priority desc, then FIFO by enqueued_at). It does NOT change item
+// state: the single-consumer model acks on completion, so a crash mid-
+// iteration simply leaves the item pending for the next drain
+// (at-least-once). limit <= 0 is treated as 1.
+func (s *Store) Peek(consumerLoop string, limit int) ([]Item, error) {
```
thane-developer: [P2] attempts is never advanced and there is no claim/visibility/dead-letter path, so a poison item remains at the head forever. Because queue_pull is read-only and ordering is priority desc/FIFO, any item the archivist cannot process/ack will be returned every pass and can starve everything behind it. The v1 scope called out claim/visibility + retry -> dead-letter for this reason. Could queue_pull claim/increment attempts, or add a queue_defer/dead-letter transition after N attempts so one bad subject cannot monopolize the consumer?

internal/app/archivist_session_close_wake.go
```diff
@@ -0,0 +1,109 @@
...
+// in the archivist's durable queue, keyed dedup on "session:<id>" so the
+// same session can't pile up. It does NOT wake the archivist — the
+// archivist is a self-paced consumer that drains its queue on its own
+// cadence (issue #1024). This is wired into both the real-time
+// EndSession callback and the summarizer's backstop scan; sharing it
+// keeps the enqueue shape identical on both paths.
+//
+// Signature matches the summarizer's wake-callback type so it can be
+// registered there unchanged.
+func (a *App) enqueueSessionCloseWork(_ context.Context, sessionID, reason string) error {
```
thane-developer: [P2] This drops the ctx, so sessionCloseEnqueueTimeout does not actually bound the EndSession callback. loopqueue.Store.Enqueue uses db.Exec without context, so if SQLite is blocked or shutting down the synchronous close path can still wait despite the 2s timeout. Either thread context into Store.Enqueue via ExecContext, or remove the timeout claim. Given this runs from EndSession, ExecContext seems like the safer fix.

thane-developer (commented):
Review comments from Codex.

internal/platform/config/config.go
```diff
@@ -315,14 +315,14 @@ type Config struct {
...
-	// documents keyed by subject. Unlike the Go-side session
-	// summarizer worker (event-triggered on session close), the
-	// curator picks its own targets each iteration. Maintains
-	// core/curator.md as its working state.
-	Curator CuratorConfig `yaml:"curator"`
+	// documents keyed by subject. It drains a durable work queue that
+	// producers (session close, frontier expansion) enqueue into; it is
+	// never event-woken, so a burst of activity can't amplify into a
+	// burst of work. Maintains core/archivist.md as its working state.
+	Archivist ArchivistConfig `yaml:"archivist"`
```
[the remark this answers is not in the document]
thane-developer: Fixed in 97bdbba — added a `case "curator":` to `rejectRetiredKeys` that errors and points the operator at the `archivist:` rename (the field shape is unchanged), mirroring the existing `platform`→`companion` guard. Covered by a new `TestLoad_RetiredTopLevelCuratorIsRejected`.

internal/app/archivist_session_close_wake.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	raw, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("marshal session-close payload: %w", err)
+	}
+	if err := a.loopQueue.Enqueue(archivist.DefinitionName, "session:"+sessionID, 0, raw); err != nil {
+		return fmt.Errorf("enqueue session-close work: %w", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
thane-developer: Fixed in 97bdbba — `Store.Enqueue/Peek/Ack/PendingCount` now take a `context.Context` and use `ExecContext`/`QueryContext`/`QueryRowContext`, and `enqueueSessionCloseWork` passes its `ctx` through instead of discarding it, so the 2s session-close enqueue timeout is actually honored. Threaded `ctx` through the `queue_pull`/`queue_ack`/`queue_enqueue` runtime-tool handlers and the tests as well.

internal/app/archivist_session_close_wake.go
thane-developer: Fixed in 97bdbba — same finding as the Copilot review. `Store.Enqueue` now uses `ExecContext` (as do `Peek`/`Ack`/`PendingCount`), and `enqueueSessionCloseWork` threads its `ctx` through, so `sessionCloseEnqueueTimeout` actually bounds the synchronous EndSession path now.

internal/state/loopqueue/queue_store.go
thane-developer: Deferring to v2, tracked in #1024. Rationale: this unbounded-growth mode is categorically milder than the incident this PR fixes. The original wake-storm was unbounded *concurrent LLM calls* saturating the model fleet; an oversized `loop_queue` is bounded disk drained by a *single, zero-spawn-rights* consumer on its own cadence — it cannot saturate the fleet. A transactional per-consumer depth cap on the new-key path is the right v2 shape and pairs naturally with the claim/visibility work in the sibling thread. Keeping v1 minimal per the plan so the cap can be tuned against real drain behavior rather than guessed.

internal/state/loopqueue/queue_store.go
thane-developer: Deferring to v2, tracked in #1024. Claim/visibility + retry→dead-letter was an explicit v1 scope cut — `Peek` is deliberately read-only (single consumer + ack-on-done = clean at-least-once). In practice the poison-item risk is softened because the consumer is an LLM agent, not a dumb worker: it can `queue_ack` an item it decides it cannot process, so a bad item is not structurally stuck at the head. Doing this properly needs a claim/last-pulled timestamp so `attempts` reflects real failed passes rather than batch under-drain (a crude attempts++-on-every-pull would over-count) — that is the v2 work, logged alongside the depth cap.

Issue #1173, opened 2026-07-02
Loop container/placement tool outputs are untyped maps; give them struct schemas


thane-developer:
Surfaced by the v0.10.1 pre-release model-facing audit. `loop_containers` builds its container rows as `[]map[string]any` (internal/tools/loop_containers_tool.go:83-86) and the placement advisory assembles as `map[string]any` (internal/tools/loop_placement_advisory.go:67-73). Both are model-facing tool outputs, and the doctrine wants generated state as typed JSON with a stable, declared schema — untyped maps make field names and presence a runtime accident rather than a compile-time contract, and they drift silently as the code evolves.

The fix is mechanical: define `LoopContainerView` and `LoopPlacementAdvisory` structs (with Godoc meeting the release-checklist standard), marshal those, and keep field names byte-identical to today's output so nothing model-visible changes at introduction. Not release-blocking for v0.10.1 because the field names are stable within the release; it becomes load-bearing the first time someone edits those maps without noticing they're a schema.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1202, opened 2026-07-03
fix(docs): reject confusable doc_write/doc_edit parameters, guard bodiless creates

### LINKED ISSUE #1201: doc_write silently eats content sent in doc_edit's vocabulary — three prod dossiers written empty
The operator noticed a suspicious commit in the kb docroot's git history: `doc_write kb:dossiers/entity-binary_sensor-zone25_garage_bay_3.md`, landing a file with nothing in it but `created` and `updated` frontmatter. The immediate fear was data loss in the document layer. The truth is better and worse: no existing data was destroyed — every one of these commits is a `new file` — but the model's *newly authored* content evaporated in flight, three times in one day, and the tool told the model everything went fine.

The forensics were conclusive. Prod's `tool_calls` log holds the exact arguments the archivist loop (running `gpt-oss:120b`) sent, and they contain a complete, well-crafted dossier — summary, claims with evidence refs, open questions, connections. It arrived in a parameter named `content`, alongside `mode: "replace_body"`. Those are `doc_edit`'s parameters. `doc_write`'s body parameter is `body`. The tool silently ignored the keys it didn't recognize, treated the absent `body` as "nothing to write", created a frontmatter-only document, and returned success. The model, told it succeeded, narrated a faithful completion: "the session was read, folded into a new dossier, and written." Three dossiers died this way on 2026-07-02 (17:03, 17:04, 22:25 CDT) while three sibling writes that happened to say `body` landed intact. The lost content was recovered from the tool-call log.

This is the same failure class as the `parent_name`-in-`spec.metadata` incident (#1103): the model puts real data where the runtime doesn't read, and the runtime's tolerance converts a naming mistake into a phantom success. The twist here is that we set the trap ourselves — `doc_write` takes `body` while its sibling `doc_edit` takes `content` for the same concept, and a local model blending the two vocabularies is not doing anything unreasonable. The tool's job at that point is to fail loudly, not to write an empty file.

Three guards close it, all shipped in the fixing PR:

- **`doc_write` rejects `content` and `mode`** with a redirect that teaches the correct parameter and names `doc_edit` for mode-based edits — a tool error puts the correction in the model's context, so the retry self-corrects.
- **`doc_edit` rejects `body`** symmetrically.
- **Creating a new document requires `body`.** "Omit to preserve the existing body" is meaningless on a create; a bodiless create is the signature of arguments that missed the schema, whatever the misnamed key was. An explicit `body: ""` remains the documented way to create a blank document, and a journal-entry-only write remains the one legitimate bodiless create.

Worth considering as a follow-up (not done in the fixing PR): a general strict-arguments posture for high-consequence mutation tools — reject any unrecognized key rather than enumerating known confusables. That's a policy decision with compatibility surface (some models attach benign extra keys), so it deserves its own discussion rather than riding a bug fix.

## Acceptance

- [x] The exact prod argument shape (`content` + `mode` on `doc_write`) errors with a teaching redirect and creates nothing
- [x] `doc_edit` with `body` errors symmetrically
- [x] Bodiless create errors; explicit `body: ""` and journal-only creates still work
- [x] Lost dossier content recovered from the tool-call log and returned to the operator

Refs #1103 (failure class), #1173 (tool-schema hygiene).

thane-developer:
## Summary

Closes #1201 — the prod incident where `doc_write` silently ate three dossiers' content. Full forensics in the issue; the short version: the archivist model sent a complete dossier in `content` + `mode: "replace_body"` (doc_edit's vocabulary), `doc_write` ignored the unknown keys, created a frontmatter-only file, and **returned success** — so the model narrated a completion that never happened. No pre-existing data was touched (every empty commit was a new file), and the lost content was recovered from the `tool_calls` log.

The trap was ours: `doc_write` takes `body` while its sibling `doc_edit` takes `content` for the same concept. A local model blending the two vocabularies isn't unreasonable — the tool's job at that moment is to fail loudly, not write an empty file. Same failure class as #1103's `parent_name`-in-metadata: real data lands where the runtime doesn't read, and tolerance converts a naming slip into phantom success.

## The guards

- **`doc_write` rejects `content` and `mode`** with redirects that teach the correct parameter and name `doc_edit` for mode-based edits. Tool errors land in the model's context, so the retry self-corrects — exactly how the #1189 phantom-success guard works for unknown service targets.
- **`doc_edit` rejects `body`** symmetrically.
- **Creating a *new* document requires `body`** (store-level, `Store.Write`): "omit to preserve the existing body" is meaningless on a create, and a bodiless create is the signature of arguments that missed the schema — this catches *any* misnamed body key, not just the two we know about. Explicit `body: ""` still creates a blank document (the documented contract); a journal-entry-only write remains the one legitimate bodiless create. All non-tool `Write` callers pass `Body` explicitly (verified: loop-create scaffold, loop outputs, intake).

The `body` parameter description now also states that creating a new document requires it, and names the `body`-vs-`content` distinction outright.

## Round 2: remove the trap, don't just fence it

The operator asked the right question: why keep two names at all? This PR now **unifies the vocabulary** — `doc_edit` and the loop-declared replace-output runtime tool take `body`, matching `doc_write`, `doc_intake`, `doc_read`'s output shape, and `doc_edit`'s own mode names (`replace_body`, `append_body`). Every model-facing markdown parameter on the document surface shares one name; journal entries stay `entry` (a genuinely distinct concept).

The legacy `content` key gets a rename-teaching error on both tools instead of silent ignoring — a model replaying old conversation history would otherwise apply an edit with empty text, the exact bug class this PR closes. And the smoking gun for how the model learned the blend: `talents/documents.md`'s own `doc_edit` example taught `content`. Updated.

The original guards remain as defense-in-depth: `mode` on `doc_write` still redirects to `doc_edit`, and the store-level bodiless-create guard catches any *future* vocabulary drift regardless of key name.

## Test plan

- [x] The exact prod argument shape (`content` + `mode`) errors with a teaching redirect and creates nothing
- [x] `mode` on doc_write redirects to doc_edit even alongside a valid `body`
- [x] `doc_edit` with `body` errors symmetrically
- [x] Bodiless create errors and creates nothing; explicit `body: ""` create and journal-only create still work
- [x] Pre-existing preserve/clear-by-intent and journal tests still green
- [x] `doc_edit` accepts `body` and rejects legacy `content` with a rename-teaching error that mutates nothing
- [x] Loop replace-output runtime tool renamed with the same guard
- [x] `just ci` green locally (unpiped, real exit code verified)

Refs #1103, #1173.

### DIFF
--- a/internal/app/loop_outputs.go
+++ b/internal/app/loop_outputs.go
@@ -87,17 +87,23 @@ func buildLoopOutputTools(store *documents.Store, outputs []looppkg.OutputSpec) 
 				Parameters: map[string]any{
 					"type": "object",
 					"properties": map[string]any{
-						"content": map[string]any{
+						"body": map[string]any{
 							"type":        "string",
 							"description": "Complete markdown body for this output. This replaces the document body as the current authoritative state.",
 						},
 					},
-					"required": []string{"content"},
+					"required": []string{"body"},
 				},
 				Handler: func(ctx context.Context, args map[string]any) (string, error) {
-					content, _ := args["content"].(string)
+					// Rename guard: this tool's parameter was unified with the
+					// document tools' body. A loop mid-conversation may replay
+					// the old key from its own history — teach, don't eat.
+					if _, hasContent := args["content"]; hasContent {
+						return "", fmt.Errorf("this tool's markdown parameter is %q (renamed from %q for consistency with the doc tools) — re-call with body", "body", "content")
+					}
+					content, _ := args["body"].(string)
 					if strings.TrimSpace(content) == "" {
-						return "", fmt.Errorf("content is required")
+						return "", fmt.Errorf("body is required")
 					}
 					result, err := store.Write(ctx, documents.WriteArgs{
 						Ref:  output.Ref,
--- a/internal/app/loop_outputs_test.go
+++ b/internal/app/loop_outputs_test.go
@@ -62,7 +62,7 @@ func TestHydrateLoopOutputsBuildsScopedToolsAndContext(t *testing.T) {
 	}
 
 	_, err = hydrated.RuntimeTools[0].Handler(context.Background(), map[string]any{
-		"content": "## Current Sense\n\nEverything is calm.",
+		"body": "## Current Sense\n\nEverything is calm.",
 	})
 	if err != nil {
 		t.Fatalf("replace output handler: %v", err)
--- a/internal/state/documents/intake_tools.go
+++ b/internal/state/documents/intake_tools.go
@@ -109,7 +109,7 @@ func (t *Tools) Commit(ctx context.Context, args CommitArgs) (string, error) {
 		out, err = t.store.Edit(ctx, EditArgs{
 			Ref:     ref,
 			Mode:    mode,
-			Content: body,
+			Body:    body,
 			Section: section,
 			Heading: heading,
 		})
--- a/internal/state/documents/mutate.go
+++ b/internal/state/documents/mutate.go
@@ -131,6 +131,16 @@ func (s *Store) Write(ctx context.Context, args WriteArgs) (*MutationResult, err
 		existingRecord = record
 	}
 
+	// A new document with no body at all is almost never intent — it is
+	// the signature of arguments that missed the schema (a misnamed body
+	// parameter is silently absent here). "Omit to preserve the existing
+	// body" is meaningless on a create, so require the body explicitly;
+	// an empty string remains the documented way to create one blank. A
+	// journal-entry-only write is the one legitimate bodiless create.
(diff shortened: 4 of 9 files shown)


nugget (dismissed):

nugget (dismissed):

thane-developer:
CI note: the `test` failures on this PR are the pre-existing `TestSubjectContextProvider_GetContext/maxFacts_limits_output` flake — code this PR doesn't touch. It fired twice in a row (evicting `fact2` then `fact1` — opposite sides of the same second-boundary straddle), so CI runners hit it routinely. Root cause + fix in #1203; cherry-picked onto this branch (`08027bd9`) so this PR's CI is self-sufficient. The patch dedupes when #1203 merges.

Issue #1221, opened 2026-07-06
Mid-turn input merge: a PullInput engine seam delivers new messages into a live turn


thane-developer:
**Phase 2 of the mid-turn message delivery arc.** Depends on the Phase 1 mailbox (#1220). Origin: design discussion 2026-07-05.

## The gap is awareness while composing, not delivery latency

When a message arrives for a loop in `StateProcessing`, the enqueue path drops a token into `wakeCh` (`internal/runtime/loop/signals.go:249-261`); the token stays buffered, so the loop re-wakes the moment the turn ends. Post-turn delivery is already near-zero. What is missing is awareness during the turn: a long tool-heavy turn composing a reply to message N is blind to N+1 and N+2, so the model ships answers that are stale or superseded, or does redundant work a later message cancels. This phase merges newly-arrived conversation content into the live turn at iteration boundaries — the same shape Claude Code uses for queued user input at tool boundaries.

## One seam, in the engine's existing idiom

Add an engine config callback in the `CheckBudget` mold (`internal/runtime/iterate/config.go:100-103`): `PullInput func(ctx) []llm.Message`, polled at the top of each iteration — strictly after tool results are appended (`internal/runtime/iterate/engine.go:346`) and before the next `ChatStream` call, so a tool_use/tool_result pair is never split. Non-empty result → append as user-role messages and continue the loop.

The plumbing is proven: `NudgeOnEmpty` already injects a synthetic user-role message mid-turn and re-loops (`internal/runtime/iterate/engine.go:401-412`), so the provider adapters handle mid-turn user messages end-to-end. Tail-appending is prompt-cache-friendly. The engine stays bus-agnostic — it calls a callback; the loop wires it. That is the entire engine diff.

Honest limitation, worth keeping in the docs: the injection window exists only between model calls. A zero-tool-call conversational reply has no boundary to inject at — and doesn't need one, because it is short and post-turn re-wake is immediate. The feature pays off exactly on long tool-heavy turns where arrival windows are wide.

## Loop wiring and provenance

The loop wires `PullInput` to the mailbox's incremental drain, reusing Phase 1's `renderEnvelope`. Rendering must carry provenance: injected content is sender-controlled text arriving while the model is mid-work — render it explicitly as channel content ("New message from Alice, arrived while you were working: …"), never in a system voice. Record each drained envelope in the conversation store at injection time so the store matches the engine's messages array (user 1 → work → user 2 → assistant).

## Guardrails

[...]

PR #1223, opened 2026-07-06
feat: add durable loop input mailbox

### LINKED ISSUE #1220: Loop input mailbox: a durable data-plane inbox for conversational loops, replacing Signal's per-sender channel
**Phase 1 of the mid-turn message delivery arc.** Origin: design discussion 2026-07-05 on notifying a running loop that more messages are enqueued for delivery. This phase is the load-bearing one and is independently valuable: it fixes two latent bugs and a structural blindness even if the later phases never ship.

## The per-sender Go channel is the wrong primitive

Signal conversation content does not ride the loop notify queue. Inbound envelopes are dispatched into a per-sender buffered Go channel (`senderChanSize = 4`) created by `ensureSenderLoop`, and the per-sender child loop's `WaitFunc` pulls exactly one envelope per turn (`internal/channels/messaging/signal/bridge.go:457`). This shape has three defects.

**Head-of-line blocking.** `enqueueSenderEnvelope` (`internal/channels/messaging/signal/bridge.go:378`) blocks on a full channel, and it is called from the single shared dispatch path that consumes signal-cli's message stream. A 5+ message burst from one sender whose loop is mid-turn stalls Signal receive for every sender until that loop drains an envelope.

**A durability hole.** The read receipt is sent after a successful enqueue into the in-memory channel (`internal/channels/messaging/signal/bridge.go:361-373`). A crash between enqueue and drain loses the message — but the sender's Signal client already shows it as read. We acknowledge mail we can lose.

**Invisibility.** The channel is bridge-private. The loop substrate cannot peek its depth, nothing can prioritize or coalesce it, and no engine feature can ever drain it. Any "N messages waiting" signal, batch turn, or mid-turn merge trips over this first.

## A data-plane mailbox belongs in the loop substrate

Give `Loop` a first-class input mailbox, distinct from `pendingNotifies`. The distinction matters: `pendingNotifies` (`internal/runtime/loop/signals.go:221`, bounded at 8, drained by `consumePendingNotifies` at `internal/runtime/loop/signals.go:315`) is control-plane — wake reasons, supervisor forcing, `loop_wake` one-shots (PR #1107). The mailbox is data-plane: actual conversation content awaiting delivery into turns. Keep them separate; notifies render in the wake summary, mailbox content renders as user messages.

Mailbox semantics: per-loop, peekable depth, drain-all at wake plus an incremental drain API (the seam Phase 2 consumes), and wake-on-nonempty riding the existing `wakeCh` coalescing (`internal/runtime/loop/signals.go:249-261`, non-blocking token send). The buffered token already means a turn that ends with a non-empty mailbox re-wakes immediately — that stays the delivery path for anything a turn doesn't consume.

Durability: ride the loopqueue chassis from #1208. `loopqueue.Store` with per-partition serialization and Peek/ack semantics already exists and drives MQTT and subscription wakes via `queuedWakeDispatcher` (`internal/app/queue_wake_dispatch.go:35-90`). Partition per loop, ack when the envelope enters a turn — "read receipt sent" and "durably queued" become the same event, closing the durability hole. Design the mailbox as the same primitive as (or a sibling of) the generic work_queue direction in #1086, not a Signal one-off.

## Signal becomes a pure protocol adapter

`dispatch()` keeps what it is good at — filtering, rate limits, binding resolution, reaction bookkeeping — but ends in a non-blocking mailbox enqueue instead of a blocking channel send. The per-sender child loop survives: one conversation, one loop, one context is the right serialization. Its bespoke `WaitFunc` closure over a private channel is replaced by the substrate's wait-on-mailbox, which also makes the conversational-loop shape reusable by construction (companion WS, any future chat surface) instead of a Signal-specific dance.

The refactor with real work in it is splitting `prepareSignalTurn` (TurnBuilder wiring at `internal/channels/messaging/signal/bridge.go:468-476`). Today it is "one envelope → whole turn." Separate the turn scaffolding (system prompt, conversation context, memory) from a reusable `renderEnvelope(env) → message` callable at turn start for a batch and — in Phase 2 — mid-turn for increments. Hang per-envelope side effects (receipt ack, `lastInboundTS`, memory attribution) off explicit lifecycle points (enqueued / delivered-into-turn) instead of pipeline stages that assume one-envelope-per-turn.

Batch turns need deliberate reply semantics: when one turn answers three messages, decide which timestamp gets the receipt, how quotes and reactions target individual messages, and how the conversation store interleaves — record each drained envelope as its own user message so the store matches the engine's messages array.

## What this fixes on day one

Head-of-line blocking is gone (dispatch never blocks). The receipt durability hole is closed (durable before ack). A burst of four messages arriving between turns becomes one coherent multi-message turn instead of four sequential blind ones. And the substrate now owns conversational-loop mechanics.

## Scope, non-goals, acceptance

No engine changes in this phase — the `PullInput` seam and mid-turn merge are Phase 2. Control-plane notifies (`pendingNotifies`, `loop_wake`, supervisor forcing) are untouched.

- [ ] Mailbox primitive in the loop substrate: durable (loopqueue partition per loop), peekable depth, drain-all and incremental drain, wake-on-nonempty via the existing wakeCh.
- [ ] Signal dispatch enqueues non-blocking; per-sender Go channels and `senderChanSize` deleted.
- [ ] Read receipt ack happens at (or after) durable enqueue; a crash between enqueue and turn no longer loses acknowledged messages.
- [ ] `prepareSignalTurn` split into scaffolding + `renderEnvelope`; wake drains all available envelopes into one multi-message turn.
- [ ] Conversation store records each drained envelope individually; receipt/quote semantics for batch turns decided and tested.
- [ ] Model-facing teaching for the batch-turn rendering per docs/model-facing-context.md and docs/model-facing-tools.md.

## Sequencing

Phase 1 (this issue) → Phase 2 #1221 (PullInput engine seam, mid-turn merge) → Phase 3 #1222 (model teaching + prod tuning). Phase 1 stands alone: its fixes are worth shipping even if the later phases are dropped.

thane-developer:
## Summary

- Add a durable loop mailbox backed by `loopqueue` for data-plane inputs, with append-only enqueue, peekable depth, drain-all wake handling, and limited incremental drain.
- Route mailbox items through loop `TurnInput` and migrate Signal per-sender loops away from private buffered Go channels.
- Render drained Signal batches as separate user messages in one loop turn, and teach the Signal channel context how to interpret batched inbound messages.

## Why

Signal dispatch previously wrote conversation content into a bridge-private buffered Go channel. A burst from one sender could block the shared Signal receive path, read receipts were sent after enqueueing into volatile memory, and the loop substrate had no visibility into pending conversation data.

This makes Signal a protocol adapter over the loop runtime mailbox: dispatch durably enqueues before receipts, child loops wake through the existing coalesced wake channel, and the loop runtime owns the mailbox mechanics.

Closes #1220

## Validation

- `just test`
- `just ci`


internal/runtime/loop/mailbox.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+	if len(stored) == 0 {
+		return nil, nil
+	}
+	items := make([]MailboxItem, 0, len(stored))
+	for _, it := range stored {
+		if err := m.store.Ack(ctx, loopName, it.DedupKey); err != nil {
```
thane-developer: **[CONFIRMED] At-most-once: items are acked (deleted) before the turn runs.**

`Ack` is an unconditional `DELETE FROM loop_queue` (queue_store.go:209-211), and this loop runs inside the wake-time drain — before `buildAgentTurn`/`runAgentTurn` (loop.go:1439 → 1587/1617). Any TurnBuilder error, runner/LLM failure, nil-turn result, or process crash after the drain permanently loses the whole batch, even though read receipts were already sent at enqueue. The iteration error path only logs and backs off; there is no requeue anywhere. This inverts the store's own documented contract (`Peek` godoc: "acks on completion … at-least-once", queue_store.go:127-130) and the #1220 acceptance criterion (ack when the envelope enters a turn).

Sub-bug on the same lines: a mid-loop `Ack` failure returns `(nil, err)` after earlier iterations already deleted their rows — those items are destroyed without ever being returned to any caller, no crash required (one transient SQLITE_BUSY suffices).

Fix shape: drain = peek only; ack after the turn commits. That is also the lifecycle Phase 2 (#1221)'s incremental mid-turn drain needs, so moving it now avoids reworking the seam later.

internal/runtime/loop/loop.go
```diff
@@ -1232,6 +1235,8 @@ func (l *Loop) run(ctx context.Context) {
 		},
 	})
 
+	l.wakeIfMailboxNonEmpty(ctx)
```
thane-developer, line 1238: **[CONFIRMED] Persisted rows are not delivered after a restart — the headline durability doesn't land.**

This check runs once per loop start, but Signal child loops are spawned lazily by `ensureSenderLoop`, only on fresh inbound traffic — `Register` never sweeps loop_queue partitions. A message accepted just before a crash (already durably enqueued and read-receipted) sits undelivered until that sender happens to message again; for a sender who never writes again, forever.

Aggravating: the partition key is the spawn-time loop *name*, whose basis changes when a contact binding appears between restarts (`signal/<phone>` → `signal/<shortContactID>`, bridge.go:432-442) — the old partition is then stranded even when the sender does write again. The store cannot enumerate partitions (no `SELECT DISTINCT consumer_loop` exists), and `Ack` is the only row-removal path, so orphaned rows accumulate in thane.db indefinitely.

Needs a boot-time sweep (enumerate pending partitions at `Register` and respawn/drain them — the archivist's boot-spawned fixed-name consumer is the in-repo precedent) plus either a binding-independent partition key or a rename/migration step when the name basis changes.

internal/channels/messaging/signal/bridge.go
```diff
@@ -379,41 +378,49 @@ func (b *Bridge) enqueueSenderEnvelope(ctx context.Context, env *Envelope) bool
...
+	payload, err := json.Marshal(env)
+	if err != nil {
+		b.logger.Warn("signal envelope marshal failed, dropping message",
+			"sender", env.Source,
+			"error", err,
+		)
+		return false
+	}
+	if _, err := b.registry.EnqueueMailbox(ctx, loopID, "signal", payload); err != nil {
+		b.logger.Warn("signal mailbox enqueue failed, dropping message",
```
thane-developer, line 414: **[CONFIRMED] A stopped child loop permanently mutes its sender.**

`senderLoops` is never evicted (four references, zero `delete`s), `ensureSenderLoop` returns the cached ID with no liveness check (line 413), and the enqueue path requires a live loop twice over: `Registry.EnqueueMailbox` errors once the loop deregisters (registry_signal.go:29-32) and `enqueueMailbox` rejects stopped/not-started loops (mailbox.go:137-141). After a `loop_stop` — reachable from the model-facing tool layer — every message from that sender lands on this warn-and-drop line: no read receipt, no reply, no respawn, until process restart.

Durable enqueue shouldn't require a live consumer at all (archivist producers already enqueue with no loop present): persist unconditionally keyed by the durable name, wake best-effort, and evict/respawn on failure here rather than dropping.

internal/channels/messaging/signal/bridge.go
```diff
@@ -572,10 +542,59 @@ func (b *Bridge) prepareSignalTurn(ctx context.Context, env *Envelope) (*loop.Ag
...
+	msgs := make([]loop.Message, 0, len(items))
+	summary := map[string]any{
+		"sender":            sender,
+		"mailbox_items":     len(items),
+		"signal_batch_turn": len(items) > 1,
+	}
+	for _, item := range items {
+		var env Envelope
+		if err := json.Unmarshal(item.Payload, &env); err != nil {
+			return nil, fmt.Errorf("decode signal mailbox item %s: %w", item.ID, err)
```
thane-developer: **[CONFIRMED] One undecodable payload kills the whole batch.**

There is no per-item tolerance: the first `json.Unmarshal` failure aborts `prepareSignalMailboxTurn` (a `renderEnvelope` error two lines down does the same), the TurnBuilder error fails the iteration — and because the batch was already acked at drain time, every *valid* message drained in the same wake is permanently lost with it. A corrupt or schema-drifted row is precisely what a queue that survives restarts and upgrades will eventually contain; once the ack moves post-turn (as it should), this same line instead becomes a permanent redeliver-fail wedge for the sender's loop.

Payload validity is a per-item property: skip-with-warn (or dead-letter) the bad row and render the rest.

internal/runtime/loop/loop.go
```diff
@@ -1476,18 +1482,22 @@ func (l *Loop) run(ctx context.Context) {
 				"supervisor_trigger": string(supervisorTrigger),
 				"attempt":            attemptCount + 1,
 				"signal_envelopes":   len(signals),
+				"mailbox_items":      len(mailboxItems),
 			},
 		})
 		iterLog.Debug("loop iteration starting")
 
-		if l.config.Handler != nil {
+		if mailboxErr != nil {
```
thane-developer: **[CONFIRMED] A transient drain error destroys this wake's notifies and strands event-driven loops.**

By the time this error short-circuits the iteration, `consumePendingNotifies()` (line 1438) has already destructively consumed the one-shot notify envelopes (`pendingNotifies = nil`, signals.go:348) — they are never passed anywhere on this path and have no redelivery, so a momentary SQLITE_BUSY permanently drops control-plane signals that had nothing to do with the mailbox.

Second half: the error path never re-signals `wakeCh`, and event-driven loops skip the sleep/backoff phase — the loop goes straight back to `waitForWake` and blocks. The still-pending mailbox rows (safely durable, since the failure was in the read) sit undelivered until the sender's next message re-wakes the loop.

Better: on drain error, log and proceed with the turn using the signals in hand, letting the pending rows ride the next wake — a durable-queue read failure shouldn't cost the volatile inputs.

internal/channels/messaging/signal/bridge.go
```diff
@@ -434,46 +441,17 @@ func (b *Bridge) ensureSenderLoop(ctx context.Context, sender string) {
...
+		Name:      loopName,
+		Operation: loop.OperationEventDriven,
 		TurnBuilder: func(tCtx context.Context, input loop.TurnInput) (*loop.AgentTurn, error) {
-			env, ok := input.Event.(*Envelope)
-			if !ok {
-				if len(input.NotifyEnvelopes) > 0 {
-					return b.prepareLoopNotificationTurn(tCtx, sender, input.NotifyEnvelopes)
-				}
-				return nil, nil
+			if len(input.MailboxItems) > 0 {
```
thane-developer, line 496: **[CONFIRMED] Mailbox items win, notify envelopes silently vanish.**

One iteration genuinely delivers both: loop.go:1438-1439 consume notifies and drain the mailbox back-to-back into the same `TurnInput`. This branch returns the mailbox turn without ever reading `input.NotifyEnvelopes`, consumption was destructive, and the `summarizeNotifyEnvelopes` prepend lives only in `buildTaskTurn` — the default path custom TurnBuilders bypass. A `loop_wake` / core-attention notify that races a queued Signal message is consumed and rendered nowhere; `prepareLoopNotificationTurn` only runs on mailbox-empty wakes.

`prepareSignalMailboxTurn` should render the notify summary too when `input.NotifyEnvelopes` is non-empty (same `FormatNotifyEnvelopes` scaffolding the notification turn already uses).

internal/runtime/loop/loop.go
```diff
@@ -1431,6 +1436,7 @@ func (l *Loop) run(ctx context.Context) {
 		}
 
 		signals, forceSupervisor, wakeTags := l.consumePendingNotifies()
+		mailboxItems, mailboxErr := l.DrainMailbox(ctx, 0)
```
thane-developer: **[CONFIRMED] Unbounded drain into a single turn — the old backpressure has no replacement.**

`DrainMailbox(ctx, 0)` drains everything pending, and `prepareSignalMailboxTurn` renders every item as its own message with no cap (`maxNotifyEventsInSummary` applies only to notify summaries). The bound this replaces — `senderChanSize = 4` with a blocking enqueue — is gone, and the remaining inflow limiter doesn't engage by default: config.go:1732's godoc claims `rate_limit_per_minute` "Default: 10", but `applyDefaults` never sets it, so an omitted key is 0 = unlimited (bridge.go:963 treats <= 0 as allow-all). A burst of 200 messages during one long turn becomes a single request with 200 user messages — context blowup or outright rejection, which under the ack-at-drain issue then loses the entire acked batch.

A per-wake drain cap here (leave the remainder pending; the wake token re-arms via the enqueue path) bounds turn size naturally, and the rate-limit default should match its godoc either way.

internal/runtime/loop/mailbox.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (l *Loop) enqueueMailbox(ctx context.Context, keyPrefix string, payload []byte) (MailboxReceipt, error) {
+	if err := ctx.Err(); err != nil {
+		return MailboxReceipt{}, err
+	}
+
+	l.mu.Lock()
+	mailbox := l.deps.Mailbox
+	loopID := l.id
+	loopName := l.config.Name
```
thane-developer, line 173: **[PLAUSIBLE — design-level] ID-addressed API, name-keyed partition: duplicate names share one mailbox.**

`EnqueueMailbox` targets a loop ID, but the store partitions by `config.Name`. The registry does not reject duplicate names (`Register` checks only ID collisions; `GetByName` documents "If multiple loops share a name, the result is undefined"), `PeekAll` doesn't claim rows, and `Ack` is idempotent — so two live same-name loops would both drain the same partition: duplicated turns and wrong-sender attribution (the mailbox turn renders under the spawning closure's `sender`, not the drained envelope's source). Marked PLAUSIBLE because reaching two live same-name Signal loops requires a specific contact-store shape, but the state is architecturally unguarded — and name drift is already real (see the rehydration thread).

Name-as-durable-identity is the right call for restart survival; it just needs to be explicit: document the partition key as the durable identity, and have `Register` (or spawn) reject/warn on duplicate names for mailbox-enabled loops.

thane-developer (commented):
Multi-angle review of the PR head (d47be5fb): 8 findings posted inline — 7 CONFIRMED by adversarial verification, 1 marked PLAUSIBLE. Verdict: **needs changes before merge.**

The unifying theme: the ack lifecycle and the enqueue/liveness coupling sit at the wrong points, so most failure paths convert "durable" messages into silent loss. Findings 1 (ack-at-drain), 2 (no boot rehydration), 3 (liveness-gated enqueue + stale sender cache), 4 (poison batch), and 5 (drain-error handling) are facets of that one root and would largely resolve together: **peek at drain / ack after the turn commits, enqueue as a pure store write with best-effort wake, and a boot-time partition sweep.** That shape also leaves the #1221 incremental-drain seam correct instead of inheriting the loss window.

On the existing Copilot comments: the two nil-`Mailbox` panic claims are incorrect — `drain` and `depth` guard the nil receiver (`if m == nil || m.store == nil`, mailbox.go:78, :111), so a nil `Deps.Mailbox` is a silent no-op, not a panic (replied on those threads). The goroutine-per-message fallback observation is valid.

Non-blocking cleanup, flagged without threads: `EnqueueMailboxByName`, `Loop.EnqueueMailbox`, and `MailboxItemsFromContext` ship with zero callers (and the ByName variant duplicates `NotifyLoopByName`'s resolution switch verbatim); `Append` binds a Go `time.Time` where `Enqueue` uses `CURRENT_TIMESTAMP`, mixing timestamp shapes in one column and forcing the new `parseQueueTimestamp` any-switch that silently swallows parse failures; the non-blocking wake-poke idiom now has three copies (enqueueNotify, enqueueMailbox, wakeIfMailboxNonEmpty); and `enqueueMailbox` pays a `COUNT(*)` per inbound message for a `PendingItems` receipt field the bridge discards.

internal/runtime/loop/mailbox.go
```diff
@@ -0,0 +1,210 @@
...
+// DrainMailbox drains mailbox items for this loop. limit <= 0 drains
+// all currently pending items; positive limits provide the incremental
+// drain API used by future mid-turn input paths.
+func (l *Loop) DrainMailbox(ctx context.Context, limit int) ([]MailboxItem, error) {
+	l.mu.Lock()
+	mailbox := l.deps.Mailbox
+	loopName := l.config.Name
+	l.mu.Unlock()
+	return mailbox.drain(ctx, loopName, limit)
+}
```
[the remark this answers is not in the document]
thane-developer: Not a panic: `drain` guards the nil receiver — `if m == nil || m.store == nil { return nil, nil }` (mailbox.go:78-80). `l.deps.Mailbox` being nil yields a typed-nil `*Mailbox`, and Go dispatches the method fine on a nil pointer receiver, so `DrainMailbox` is a silent no-op for loops without a mailbox (which is why CI passes with every core loop unwired). The guard could arguably live in `Loop.DrainMailbox` for legibility, but there's no runtime defect here.

internal/runtime/loop/mailbox.go
```diff
@@ -0,0 +1,210 @@
...
+
+// MailboxDepth returns the number of pending mailbox items for this loop.
+func (l *Loop) MailboxDepth(ctx context.Context) (int, error) {
+	l.mu.Lock()
+	mailbox := l.deps.Mailbox
+	loopName := l.config.Name
+	l.mu.Unlock()
+	return mailbox.depth(ctx, loopName)
+}
```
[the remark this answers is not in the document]
thane-developer: Same as the DrainMailbox thread: `depth` guards the nil receiver — `if m == nil || m.store == nil { return 0, nil }` (mailbox.go:111-113) — so `MailboxDepth`/`wakeIfMailboxNonEmpty` are no-ops when `Deps.Mailbox` is nil, not panics.

internal/channels/messaging/signal/bridge.go
```diff
@@ -379,41 +378,49 @@ func (b *Bridge) enqueueSenderEnvelope(ctx context.Context, env *Envelope) bool
...
 	}
-	b.ensureSenderLoop(ctx, env.Source)
-
-	b.mu.Lock()
-	ch, ok := b.senderChans[env.Source]
-	b.mu.Unlock()
+	if b.registry == nil || b.mailbox == nil {
+		go b.handleEnvelope(ctx, env, nil)
+		return true
+	}
```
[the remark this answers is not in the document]
thane-developer: Addressed in ec9b7156: the no-mailbox Signal path now runs synchronously before the read receipt can be sent.

internal/runtime/loop/mailbox.go
thane-developer: Addressed in ec9b7156: mailbox delivery now peeks rows and only acks them after a successful or explicitly no-op turn.

internal/runtime/loop/loop.go
thane-developer: Addressed in ec9b7156: Signal registration sweeps pending signal mailbox partitions and uses a stable hashed sender partition, migrating old pending partitions when found.

internal/channels/messaging/signal/bridge.go
thane-developer: Addressed in ec9b7156: Signal enqueue persists first, evicts stale cached child loop IDs, respawns consumers, and wakes the live loop best-effort.

internal/channels/messaging/signal/bridge.go
thane-developer: Addressed in ec9b7156: malformed Signal mailbox rows are skipped with warnings so valid rows in the same batch still render.

internal/runtime/loop/loop.go
thane-developer: Addressed in ec9b7156: mailbox drain errors now warn and re-wake for retry while notification input continues through the turn.

internal/channels/messaging/signal/bridge.go
thane-developer: Addressed in ec9b7156: Signal mailbox turns now include the formatted notify summary when mailbox rows and notify envelopes arrive together.

internal/runtime/loop/loop.go
thane-developer: Addressed in ec9b7156: mailbox drains are capped per wake and the loop re-wakes when more rows remain.

internal/runtime/loop/mailbox.go
thane-developer: Addressed in ec9b7156: mailbox-enabled duplicate loop names are rejected, and Signal mailbox partitions now use stable hashed sender identity.

internal/runtime/loop/loop.go
```diff
@@ -1625,6 +1641,16 @@ func (l *Loop) run(ctx context.Context) {
 			}
 		}
 
+		if mailboxErr == nil && len(mailboxItems) > 0 && err == nil && (result != nil || noOp) {
```
thane-developer, line 1644: **Residual (follow-up scale, not a blocker): failed turns retain their items but nothing retries them on event-driven loops.**

The retention half is right — on `err != nil` this block is skipped, the rows stay pending, and `TestLoopMailboxKeepsItemsWhenTurnFails` proves redelivery. But `wakeIfMailboxNonEmpty` only runs inside this success-gated block (and on the drain-error path at ~1461), and event-driven loops skip the error-backoff sleep — after a failed iteration they return straight to `waitForWake` with the wake token already consumed. So a transient runner/LLM error (one Anthropic 500) leaves the sender's messages durably queued but undelivered until their *next* inbound message, a notify, or a restart re-wakes the loop. For a conversational channel that reads as "went silent until I double-texted."

Note the naive fix — adding `wakeIfMailboxNonEmpty` to the failure path — would hot-spin on a persistent failure (wake → fail → re-wake), bounded only by the consecutiveErrors breaker. The right shape is backoff-aware: let event-driven loops take the error-backoff sleep when mailbox items remain pending (instead of skipping sleep entirely), or schedule a delayed poke. Fine as a fast-follow issue if you'd rather keep this PR's scope closed.

thane-developer (commented):
Re-verified ec9b7156 against all 8 review threads: every fix is real and test-backed, and the resolutions are accurate — peek-only drain with post-turn ack (redelivery proven by `TestLoopMailboxKeepsItemsWhenTurnFails`), boot rehydration with old-partition migration, persist-first enqueue with cache eviction/respawn, per-item poison skip (and `turn == nil` → `noOp` → ack, so an all-skipped batch still clears — no wedge at the capped drain head), drain-error tolerance with re-wake, notify+mailbox merged turns, the 16-item wake cap with remainder re-wake, and the duplicate-name Register guard with stable hashed partitions. CI green.

One residual gap worth a follow-up (inline comment on the ack block): the wake re-arm lives inside the success-only ack path, so a *failed* turn retains its items but never retries them on event-driven loops. Not a blocker for this PR — strictly better than both the original commit and the pre-PR channel behavior — but it should land before this path sees prod traffic bursts.

internal/runtime/loop/loop.go
thane-developer: Fixed in 514cc368: a failed turn with retained mailbox rows now takes the error backoff at the bottom of the loop (floored at `eventDrivenErrorBackoff`, mirroring the wait-error path) and then re-arms the wake, so retained rows retry without waiting for the next external enqueue. A fresh enqueue during the backoff cuts the sleep short via wakeCh, and MaxIter-exhausted loops skip the backoff and exit promptly. Regression test: `TestLoopMailboxRetriesAfterFailedTurn` (fail-once-then-succeed runner; asserts self-retry with no new enqueue and depth 0 after ack). `just ci` green locally.

internal/state/loopqueue/queue_store.go
```diff
@@ -147,6 +193,81 @@ func (s *Store) Peek(ctx context.Context, consumerLoop string, limit int) ([]Ite
...
+	query := `
+		SELECT DISTINCT consumer_loop
+		FROM loop_queue
+		WHERE status = ?`
+	args := []any{StatusPending}
+	prefix = strings.TrimSpace(prefix)
+	if prefix != "" {
+		query += ` AND consumer_loop LIKE ?`
+		args = append(args, prefix+"%")
+	}
```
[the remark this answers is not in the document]
thane-developer: Fixed in b3b631dd: `PendingConsumers` now escapes LIKE wildcards in the prefix (`%`, `_`, and `\`) via a new `escapeLikePrefix` helper and adds `ESCAPE '\'` to the query, so the prefix is matched literally. No live bug today (the sole caller passes `"signal/"`), but it's a generic substrate method and future callers shouldn't be able to widen the match by accident. Regression test `TestStore_PendingConsumersEscapesLikeWildcards` (partition `signal/a_b` vs `signal/axb` — the unescaped `_` would match both).

internal/channels/messaging/signal/bridge.go
```diff
@@ -379,101 +387,109 @@ func (b *Bridge) enqueueSenderEnvelope(ctx context.Context, env *Envelope) bool
...
 	}
-	b.ensureSenderLoop(ctx, env.Source)
-
-	b.mu.Lock()
-	ch, ok := b.senderChans[env.Source]
-	b.mu.Unlock()
-	if !ok {
+	if b.registry == nil || b.mailbox == nil {
+		b.handleEnvelope(ctx, env, nil)
+		return true
```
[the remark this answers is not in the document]
thane-developer: Fixed in b3b631dd: `handleEnvelope` now returns an error (propagating both turn-prep failures and the previously-discarded runner error), and the no-mailbox fallback in `enqueueSenderEnvelope` returns false on failure so `dispatch` skips the read receipt — the legacy path no longer acks messages it couldn't process. A nil turn (nothing to answer) stays a successful no-op and still acks. Production always wires the mailbox so this path is legacy/test-only, but you're right that it reintroduced the ack-we-can-lose behavior and swallowed the runner error. Regression test `TestBridge_EnqueueWithoutMailboxGatesReceiptOnTurnOutcome` (failing runner → false, successful runner → true).

nugget (approved):

Issue #1250, opened 2026-07-24
Loop outputs as audience-scoped projections: tiered publishing, working notes, and fidelity-aware context injection


thane-developer:
## Summary

Thane's service loops are becoming domain curators — a loop that owns "the state of the office on the ranch" and keeps an understanding of it current. This issue is the contract for how that understanding reaches every consumer at the right fidelity: a declared ladder of detail levels (`status_line` → `teaser` → `digest` → `full`), a private working-notes space no consumer sees, and bindings that project one publish onto documents, Home Assistant surfaces, search results, and ambient context. One curation, many audiences, each at the cost that surface deserves.

The convention was proved by hand in production first (HUD / WHISPER / SUMMARY / DETAILS sections in maintained markdown), and it fails the way hand conventions fail: the model must remember the headings, nothing enforces the sizes, and no consumer actually picks a level. This issue makes it a contract.

## Status (re-grounded 2026-08-11) — the release line

The core is built and the scope of *this issue* is now fixed to what closes in **v0.10.3**: finish phase 2's read surfaces (teaser as the search snippet, facets advertised per hit, outputs on LoopView and `/v1` with content negotiation) and land the one phase-3 item that turned from aspiration into liability — `renderLoopOutputContextWithNow` still byte-truncates a loop's own document into its context, which facets made *worse*, since an authored projection now exists and is not being used. The rest of phase 3 (medium defaults, relevance-triggered injection, the status-line rail) lands as consumers demand it, with #1351's metacog status_line as the deliberately chosen first one.

**Phase 4 and the outward motion are split out to #1356** (v0.10.4): HA entity and MQTT bindings, the complication adoption, and facet-change wake subscriptions. The #1306 revert findings, the binding-fork resolution, and the parked #1302 slot registry transfer there as starting capital. This issue closes on in-document fidelity: publish, store, read, inject.

## The customer walkthrough

One loop curates the office domain. From that single curation:

- **Another thane session** mentions the office in passing, and the office loop's status line or teaser appears in that session's related context — matched by search against the incoming request, injected at a fidelity proportional to the match. Nobody subscribed to anything; relevance did the routing.
- **A kiosk in physical space** shows a digest-length markdown card, bound through a Home Assistant entity. The wall gets the two-kilobyte version, always current.
- **An explicit search** returns the authored teaser as the snippet, with the ref as the link deeper and the available facets advertised so the next step is one deliberate call.
[...]

PR #1253, opened 2026-07-26
feat(loops): add structured payload output tier with rendering targets

thane-developer:
## Summary

Loop outputs had two tiers and both were documents — `maintained_document` asks the model for a complete rewrite, `journal_document` for an appended entry, and both hand it a free-text body. That is the wrong shape for a *display*. An Apple Watch complication has four lines of roughly twenty characters, a gauge that wants a number in `[0,1]`, and no capacity for a paragraph that ran a little long.

This adds `structured_payload` as a third tier, built so it is one of many rather than a one-off.

A declaration names a **target** from a new registry (`internal/model/outputtargets`) describing one renderable surface: named slots, their kinds, their rune budgets, and which slot is primary. That single definition becomes three things at once:

- the JSON schema the generated `set_output_*` tool advertises, so the model sees `title (string, max 18 characters)` rather than a free-text field;
- the validator the handler runs;
- the slot contract rendered into the loop's context each iteration, alongside the last published payload.

**Validation rejects rather than repairs.** An over-budget string on a watch face is an unreadable fragment with nothing to signal that anything was dropped, so the handler returns an error naming the slot and the limit. Only whitespace and color-hex casing are canonicalized. Every call replaces the whole payload — a surface with no history has no meaningful partial update.

**Payloads publish over MQTT as Home Assistant discovery sensors:** the primary slot becomes the state, the rest become JSON attributes, which keeps the downstream binding trivial. An entity-sourced complication then refreshes on the watch itself without the phone in range.

Two targets ship — `apple_watch.rectangular` and `apple_watch.circular`. Adding a third is adding a `Target` value; the loop layer, tool generation, and sink are unchanged.

```yaml
outputs:
  - name: watch_status
    type: structured_payload
    target: apple_watch.rectangular
    ref: mqtt:watch_status      # → sensor.<device>_watch_status
```

Supporting changes: `mqtt.Publisher` gains `EnsureSensor` for runtime discovery registration (loops hydrate long after `Start`, where `RegisterSensors` no longer takes effect), and `ObjectIDPrefix` is exported as a function so an entity ID can be predicted before a publisher exists.

Teaching pass per `docs/model-facing-tools.md`: the loop spec output schema, the `loops` talent, the `agent-loop` and `document-roots` docs that said only two shapes exist, the tool reference, and a new `docs/understanding/defined-outputs.md`.

## Test Plan

- [x] `just ci` passes locally
- [x] New/changed behavior has test coverage
- [x] Manual verification (describe what you checked):

`just ci` is clean through `fmt-check`, `mod-tidy-check`, `config-generate-check`, `architecture-check`, `lint` (0 issues), and `lint-openapi`. `link-check` skipped locally (lychee not on PATH; CI enforces it).

`just test` leaves 24 failures, all of which reproduce identically on a clean `main` checkout — commit-signing tests (SSH allowed-signers not configured in this container) and one root-user permission test (`TestStoreMoveRestoresOverwrittenDestinationWhenSourceRemovalFails`, where a chmod-denied directory does not block root). Verified by diffing the failure sets with and without this branch: identical, 24 vs 24. This change introduces no new failures.

New coverage:

- `internal/model/outputtargets` — registry coherence (every target has a summary, binding text, exactly one required text primary slot, lower_snake_case slot names), `validate` rejection table, schema shape and advisory-ness, `Normalize` success and teaching-error tables including multibyte budgets, `json.Number`/string/int fraction coercion, explicit-null clearing, and runnable `Example` tests.
- `internal/runtime/loop` — `structured_payload` validation: mode/type pairing, missing and unknown target, target on a document output, document ref for a structured payload, entity-suffix grammar, and JSON round-tripping of the new `target` field.
- `internal/app` — tool generation against a fake sink: schema advertises every slot, payload normalization reaches the sink canonicalized, rejected payloads never reach the sink, sink failures surface to the model, hydration fails without MQTT configured, and the context block carries the contract plus last-published state and freshness delta.

## Checklist

- [x] Commits are signed
- [x] Commit messages use conventional format (`feat:`, `fix:`, etc.)
- [ ] Related issue referenced (`Refs #NNN` or `Closes #NNN`) — no existing issue found for this work
- [x] Impacted documentation updated (`docs/`, `AGENTS.md`, config examples, CLI help)
- [x] No new third-party dependencies without discussion

## Notes for review

- **Colors are best-effort by design.** An entity-sourced complication cannot template its colors — only a template-sourced one can, reading the same attributes. And watchOS renders on-face complications in the face's accent palette, so expect full fidelity in the Smart Stack and a tinted approximation on the face itself. The slot descriptions say this; the gauge *fill level* is the reliable on-face signal.
- **The last-published record is in-process only.** MQTT retains the payload on the broker, so a restarted Thane has a display it has not written to yet. The context block says so explicitly rather than leaving the absence ambiguous.
- **Rune budgets are deliberately tighter than what the widget accepts.** The `accessoryRectangular` slot on the Modular Ultra face clips earlier than on other faces, so the budgets target "fits everywhere" over "fits the roomiest face."
- `scripts/architecture.baseline` bumped for the new package (64→65) and the sink interface (84→85).

---
_Generated by [Claude Code](https://claude.ai/code/session_01BW8rCYUvE4PgnMAupvxeXb)_

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,6 +30,8 @@ complete onboarding guide for Home Assistant users.
   prompt is assembled from persona, talents, knowledge, and session state
 - **[Document Roots](understanding/document-roots.md)** — How `paths:`
   becomes named local document collections that Thane can keep track of
+- **[Defined Outputs](understanding/defined-outputs.md)** — How a loop
+  declares what it owns, and the tiers between a document and a display
 - **[Trust Architecture](understanding/trust-architecture.md)** — Safety
   through structural enforcement, not prompt compliance
 - **[Glossary](understanding/glossary.md)** — Canonical definitions for
--- a/docs/reference/tools.md
+++ b/docs/reference/tools.md
@@ -238,13 +238,20 @@ or signed deltas like `-604800s`.
 | `doc_move_section` | Move one named section into another document. |
 | `doc_journal_update` | Append or update a journal-style entry. |
 
-Loop-declared document output tools are request-scoped and do not appear
-in the global catalog above. When a loop declares a maintained document
-or journal document output, Thane generates tools such as
+Loop-declared output tools are request-scoped and do not appear in the
+global catalog above. When a loop declares a maintained document or
+journal document output, Thane generates tools such as
 `replace_output_metacognitive_state` or `append_output_daily_notes` only
 for that loop run. These tools route through managed document roots, so
 root policy, indexing, and provenance remain centralized instead of
 being reimplemented in each loop prompt.
+
+A loop can also declare a `structured_payload` output naming a rendering
+target, which generates a `set_output_<name>` tool whose schema is that
+target's slot contract — named slots with types and character budgets
+instead of a free-text body. Those payloads publish as Home Assistant
+entities rather than documents. See
+[Defined Outputs](../understanding/defined-outputs.md).
 
 ## `email` — inbox traffic
 
--- a/docs/understanding/agent-loop.md
+++ b/docs/understanding/agent-loop.md
@@ -111,19 +111,29 @@ definition. A declared output is an explicit contract: this loop is
 responsible for maintaining or appending to this document, and Thane
 generates the narrow tool surface for that job at runtime.
 
-Two output shapes exist today:
+Three output shapes exist today:
 
 - **Maintained document** outputs replace the whole current document
   through a generated `replace_output_<name>` tool.
 - **Journal document** outputs append a timestamped entry through a
   generated `append_output_<name>` tool.
+- **Structured payload** outputs fill the named slots of a rendering
+  target — an Apple Watch complication, say — through a generated
+  `set_output_<name>` tool whose schema is that target's slot contract.
+  The payload publishes as a Home Assistant entity rather than a
+  document.
 
 The same declaration also feeds context assembly. Each turn receives a
-`Declared Durable Outputs` block with the output name, document root
-reference, generated tool name, current content or recent journal tail,
-and any truncation markers. The model should use those generated tools
-instead of generic file tools; the document root owns path safety,
-indexing, provenance, and signature policy.
+`Declared Durable Outputs` block with the output name, destination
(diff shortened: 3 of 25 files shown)


thane-developer:
Closing as superseded, with the substance shipped rather than discarded.

**What landed.** #1302 lifted this branch's `internal/model/outputtargets` onto main unchanged — the whole registry, ~940 lines plus ~475 of tests: named slots, per-slot kinds and rune budgets, one designated primary, and a single definition producing the argument schema, the validator, and the contract a model reads. `EnsureSensor` came with it, because loops hydrate long after `Publisher.Start` and an entity created at hydration would otherwise wait for a reconnect to reach Home Assistant.

It transplanted with no edits at all. The registry never knew about loops — it describes surfaces and the values they hold, and who publishes to them lives elsewhere — so it imports only stdlib and passed its own tests against main on the first try.

**What did not, and why.** The framing: `structured_payload` as a third loop output type with its own `set_output_*` verb, and the wiring in `output.go`, `loop_outputs.go`, and `loop_output_structured.go`.

#1250 replaced that model in the interval. A loop's core is what it knows and thinks; facets are views cut from it for particular surfaces. Under that reading a watch complication is not a third kind of output — it is another facet of one, cut for a surface that renders no prose. The output type stays `maintained_document`, and the slot contract becomes an attribute of a facet rather than a parallel path.

Two things made that switch cheap. #1301 gave facets attributes, so a target reference is additive rather than a migration. And the files that would have conflicted with main after the facets rename and the journal retirement turned out to be exactly the files being discarded — the conflicts and the discards were the same set.

**One debt carried forward.** `Target.Schema` still describes the `set_output_*` verb from this design. It reaches no model today because nothing calls it; the consumer that wires the registry owns fixing it, and #1302 says so rather than leaving it to be discovered.

Continues in #1250. Thanks — the registry was the right abstraction, and it outlived the design it was written for.

PR #1306, opened 2026-07-27
A facet may be cut for a surface instead of a reader

thane-developer:
`status_line`, `teaser`, and `digest` are all cut for readers, differing only in how much room the reader has. A watch complication is not the next size down that ladder — it has four short lines, a gauge wanting a number between 0 and 1, and nowhere to put a sentence, so prose short enough to survive it reads badly everywhere and renders badly there. Asking a loop to write one more projection was asking the wrong question, and the answer is that a facet is a face cut for a viewing angle, not a length.

So a facet is now identified either by name or by target, and the two are alternatives rather than a name with a modifier:

```json
"facets": ["status_line", "teaser", {"target": "apple_watch.rectangular"}]
```

Everything about that surface comes from the `outputtargets` registry, which has been sitting in the tree without a caller since #1253 and is what this PR finally plugs in. The section heading, the publish argument name, the slot types and budgets, the words the model reads before it writes, and the validation its values must pass are all read from the registry — so adding a surface adds no code in the loop package at all, and the schema can never offer a target that does not exist.

## The argument is the slots, not a string

This is what declaring a target buys over declaring a bare `format: json` facet. The model fills `value`, `title`, `subtitle`, `fraction` — each typed, each carrying its own budget in the tool schema — instead of hand-encoding JSON into a text field and learning the limits by being rejected. `Target.Schema()` and `Target.Normalize()` get their first callers here, which is the shape they were written for.

`Normalize` runs on the way in, and `ValidateFacetPayload` runs it again on any payload assembled another way, so a repair, a migration, or a seeded document meets the same contract a tool call does rather than a weaker one.

## What the document stores

The canonical slot map: trimmed text, uppercased hex, sorted keys, rendered as a fenced JSON section under the surface's own heading. Slots rather than the payload a sink would derive from them — a stored payload has already thrown away which slot each value came from, and the document is the store every binding is re-seeded from after a restart.

## Three seams unified rather than duplicated

- The section table became **per-spec**. Render, parse, validate, and the tool schema still read from one source now that the table is assembled rather than fixed, which is the property that kept those four from drifting in the first place.
- `thane_loop_create`'s facets parser now decodes through `FacetSpec` instead of its own string-only reading, so the guided path accepts exactly what the raw path does. It still reports the failing element by index — decoding the array in one shot loses that, and an author reading `cannot unmarshal number` should not have to count facets by hand.
- Turning publish arguments into a payload moved from `internal/app` into the contract itself. That is what lets the talent gate below check the taught sample against the real code rather than a copy of it.

## The teaching gate

`TestTalentExampleSpecValidates` already decoded the worked example's spec. It now also decodes and validates the example's **publish sample** through the real path: an over-budget slot or a missing projection in the talent fails the build instead of teaching a call that fails. Verified by mutation in both directions.

`talents/loops.md` gains a section on facets cut for a non-prose surface, and the worked example in `talents/loops-examples.md` declares and fills one. That example's stale working-notes paragraph (still describing notes as an append-only log, and naming a `note` argument that is spelled `notes`) is corrected in passing.

## Test plan

- [x] `just ci` green
- [x] Registry validation is real: removing the `Normalize` call fails 5 of 7 subtests in `TestTargetFacetPayloadIsCheckedAgainstTheRegistry` — the other two are caught earlier by the JSON check, which is correct
- [x] Talent gate is real: an over-budget `value` slot and a removed `teaser` argument each fail `TestTalentExampleSpecValidates`
- [x] Target facets round-trip through the rendered document, including alongside reading projections
- [x] A rejected publish writes nothing
- [x] Spec serialization keeps the target (the short form is for a bare name only)
- [x] `TestNoTargetTitleCollidesWithAReservedHeading` guards the parser against a target titled like a contract heading
- [ ] Declare a target facet on a live loop and confirm the complication renders — needs the MQTT/HA binding, which is the next piece

`output_facets.go` crossed the 500-line architecture threshold, so the target seam moved to its own `target_facet.go` rather than the baseline moving.

Refs #1250

### DIFF
--- a/internal/app/loop_output_publish.go
+++ b/internal/app/loop_output_publish.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/nugget/thane-ai-agent/internal/model/outputtargets"
 	looppkg "github.com/nugget/thane-ai-agent/internal/runtime/loop"
 	"github.com/nugget/thane-ai-agent/internal/state/documents"
 )
@@ -34,14 +35,7 @@ func buildFacetPublishTool(store *documents.Store, output looppkg.OutputSpec, no
 	properties := make(map[string]any, len(fields)+1)
 	required := make([]string, 0, len(fields))
 	for _, field := range fields {
-		description := field.Guidance + looppkg.FormatGuidance(field.Format)
-		if field.MaxRunes > 0 {
-			description = fmt.Sprintf("%s Maximum %d characters.", description, field.MaxRunes)
-		}
-		properties[field.Key] = map[string]any{
-			"type":        "string",
-			"description": description,
-		}
+		properties[field.Key] = facetArgumentSchema(field)
 		required = append(required, field.Key)
 	}
 	if notes != nil {
@@ -61,7 +55,7 @@ func buildFacetPublishTool(store *documents.Store, output looppkg.OutputSpec, no
 			"required":   required,
 		},
 		Handler: func(ctx context.Context, args map[string]any) (string, error) {
-			payload, err := facetPayloadFromArgs(output, args)
+			payload, err := output.FacetPayloadFromArgs(args)
 			if err != nil {
 				return "", err
 			}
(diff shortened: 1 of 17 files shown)


thane-developer:
Parking this: it is phase 4 work landed ahead of phases 2 and 3.

The umbrella is explicit that nothing reads a published facet yet — that starts in phase 2 (read surfaces) — and that #1253's complications are phase 4, *after* fidelity-aware assembly. This PR jumped straight to the last phase, and it shows: there is no sink, so nothing here can be exercised end to end, and the design conversation it triggered surfaced that most complications want a declarative data binding rather than model-authored slots at all. Building a second mechanism for a surface we cannot yet use is the wrong order.

The registry work stands and is untouched on main. The non-target improvements from this branch (the `note`/`notes` teaching bug, the publish-sample gate, the guided/raw facet parser parity) are being split out separately, since one of them is a live bug.

Reopen against phase 4, once phase 2 gives facets a consumer.

Issue #1086, opened 2026-06-25
Unify ego/metacognitive/archivist into the loop-definition envelope (supersedes #999, #1082)


thane-developer:
## ⚠️ Status & revised direction (2026-06-25)

**Phase 0 is DONE — shipped in v0.10.0.** The agent-authored loop envelope now matches what the core loops use:
- #1087 — egress default-deny on `thane_curate` (#696)
- #1088 — enumerated loop-spec schema for the agent-facing loop tools
- #1089 — prompts → declarative `Task` + `SupervisorProfile.Instructions`
- #1090 — `subscriptions[].forecast` schema fix
- #1091 — `thane_curate` `quality_floor` / `exclude_tools` / `metadata` knobs (with fail-fast validation)

**Direction correction — config-file declaration is RETAINED.** The original plan below (D1 / "Phase 1") called for deleting `MetacognitiveConfig` / `EgoConfig` / `ArchivistConfig` and migrating the three loops to pure config-seeded definitions (config stops being source of truth). **That is dropped.** Operators keep declaring and tuning these three loops via config. The config change is instead the **#999 logic-dedupe** — one shared `ServiceLoopConfig` collapsing the 3× `applyDefaults` blocks and the 3 parallel validators, with the config keys and operator UX unchanged — landing as a **v0.10.0 close-out**. Treat D1 / "Phase 1" below as **superseded**; the resolution path is no longer struct-deletion + config-source migration.

**Remaining work (moved to v0.10.1):**
- **Frontier-spend governor (Phase 2):** gate supervisor turns for overlay / agent-authored loops before exposing the supervisor knobs on `thane_curate` (the #1024 cost class).
- **Generic work-queue (Phase 3):** generalize the archivist's loop-private durable queue so any loop can declare one. The archivist's queue-tool hook stays snowflake until this lands; the core loops remain config-declared throughout.

Reframed thesis: **enrich agent-authored loops to the envelope the core loops already use (governor + queue), and dedupe the core-loop config — while keeping the core loops config-declared.** *Not* "delete the core loops' config." Everything below is the original analysis, retained for context.

---

## Thesis

[...]

PR #1317, opened 2026-07-28
The document says where the loop belongs

thane-developer:
Follow-up to #1315, which added a warning for this. The warning is the right safety net for a hand-authored document and the wrong fix for a shipped one.

## The asymmetry

ego, metacognitive, and archivist take every field from their shipped document except one. Their parent was assigned in Go, after the spec was built, on the branch that appends a built-in:

```go
spec, err := reg.DefinitionSpec(a)
spec.ParentName = cognitionContainerName
```

A definition read from `<core>/loops` sets `hasDefinition`, never reaches that branch, and lands at the graph root. The document owned everything about the loop except where it sits.

The failure is silent, which is what makes this a breaking change rather than another guard. The loop boots, runs, and does its work. The only evidence is a cognition container that used to have three children and now has none, and nothing connects that to the file someone edited. A default the documents cannot see is a default they cannot be read against.

`parent_name: cognition` now lives in the three documents and nothing assigns it afterwards. An operator who copies a shipped document to override it gets the parent along with everything else — which is what the copy already implied.

## Breaking

Anyone who has already written a `<core>/loops` document for one of these three without a `parent_name` had that loop move to the root at the previous release, and it stays there until the field is added. `thane validate` names it, from #1315.

## The parity tests split on this field

Honestly, rather than by normalizing it away. `TestCoreLoopDocsMatchTheGoSpecs` still excludes `ParentName` — the Go builders never carried one — and now says why. `TestBuiltInSpecsNowComeFromTheShippedDocuments` asserts it positively instead of copying `got` onto `want`, because that test can actually observe whether the document's value survived. It fails when the field is stripped from a shipped document; the normalized version could not.

## Test plan

- [x] `just ci` clean, no architecture drift, generated mirror regenerated
- [x] Removing `parent_name` from a shipped document fails the boot-path parity test with the field named
- [x] A core-defined loop keeps its declared parent even though the container it names is appended after it — startup topologically orders containers before services, so slice position does not decide parenting
- [x] Production's three documents already carry the field explicitly, so this changes nothing for that instance

Refs #1086

### DIFF
--- a/internal/app/coreloop_docs_parity_test.go
+++ b/internal/app/coreloop_docs_parity_test.go
@@ -57,8 +57,10 @@ func TestCoreLoopDocsMatchTheGoSpecs(t *testing.T) {
 			if err != nil {
 				t.Fatalf("decodeCoreLoopDefinition: %v", err)
 			}
-			// ParentName is applied by the caller, not carried by either
-			// source, so it is not part of what the document owns.
+			// The document carries parent_name and the Go builder never
+			// did, so this one field is expected to differ. It is checked
+			// against the boot path in
+			// TestBuiltInSpecsNowComeFromTheShippedDocuments instead.
 			got.ParentName = tt.want.ParentName
 			if !reflect.DeepEqual(got, tt.want) {
 				t.Fatalf("document and Go spec disagree.\nfrom document: %#v\nfrom Go:       %#v", got, tt.want)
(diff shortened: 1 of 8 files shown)


thane-developer:
Pushed a fix for a case this branch opened: `builtInContainerDefinitionSpecs` gated the cognition container on **config** enablement, while `buildLoopDefinitionBaseSpecs` registers a core loop when config enables it **or** a core document defines it. A config-disabled, document-defined loop therefore named a container that was never built — and an unresolvable `parent_name` is dropped with a warning and default-parented to the root, which is the silent re-parenting this PR exists to remove, arriving through a second door.

Unreachable before this branch, since a document carried no `parent_name` at all. The gate now derives from `coreServiceLoops` so a fourth core loop is covered by existing.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nugget (approved):

Issue #1318, opened 2026-07-28
The root an operator founds is not the root an agent writes


thane-developer:
## Thesis

#1316 shapes `core` into the instance's identity evidence: a birth commit, a long-lived signing key, a channel CA, an admission result, a current HEAD, and an operator-anchored-or-self-signed posture that a remote party can eventually witness. That is a claim about who founded this Thane and who has been entitled to change it since.

Core cannot carry that claim while the agent writes to it. Not because agent writes are untrustworthy, but because they are a different kind of fact, arriving on a different schedule, entitled by a different key. Agent-produced state needs exactly the same integrity treatment core gets — signed history, admission, verification — and needs it under its own policy. So it needs its own root, held to the same standard, with a different signer set and a different authoring mode.

The default install posture we should promote is **operator-only signers on core**, with `config.yaml` shielded from agent editing. Most operators will want that, and today nothing offers it.

## The posture claim is currently true and misleading

Production's core was founded by the operator — the birth commit is `contact@nugget.info`, so admission survives narrowing the seed set. Of the commits after it, 97 of 100 are the agent's, including HEAD.

An identity endpoint reporting `operator_anchored` off the birth commit would be accurate about founding and silent about everything since. #1316's acceptance criteria explicitly guard against a misleading successful posture, and this is the shape that failure takes: every individual check passes, and the summary means less than a reader will take it to mean.

## HEAD as evidence versus HEAD as a counter

#1316 wants to report the current core HEAD and whether the worktree is clean against it. Both are useful facts about a root that changes when an operator changes something. Neither survives a root that a loop rewrites every fifteen minutes.

A churning HEAD is not a pinnable identity fact. And cleanliness against HEAD becomes a race: a validate run can land between a document write and its commit, so a check that exists to say "what is on disk is what was signed" starts producing failures that mean "you asked during a write." A check that is flaky is a check operators learn to skip, which costs more than not having it.

Under operator-only core, HEAD moves when the operator moves it. That is what makes it evidence.

## Where the line falls

Not agent-written versus operator-written — that describes the drift, not the intent. The line is **what constrains Thane versus what Thane produces**.

[...]

Issue #1316, opened 2026-07-28
feat: expose core-backed Thane identity evidence through the native API


thane-developer:
## Problem or Opportunity

Thane now establishes durable cryptographic identity in `core/`: an Ed25519 signing identity, a channel CA, a single signed birth commit, and an operator- or self-anchored trust surface. Admission and head verification can judge important parts of that history locally, but no stable native API contract lets an operator or client ask the running instance:

- which Thane identity it is presenting;
- which core birth commit that identity descends from;
- which core HEAD is currently active;
- whether birth admission and current-tree verification succeeded; or
- which public-key and channel-CA fingerprints should be pinned or compared out of band.

`GET /v1/system` exposes process health, uptime, and build information. Those are ephemeral operational facts, not durable identity evidence. Without a separate identity surface, clients cannot pin the production instance today, and future Thane-to-Thane witnessing would have no stable subject statement to sign.

The first core commit's timestamp is useful evidence but not an independently trustworthy birth date: Git timestamps and bootstrap wall-clock values are selected by the founding machine. The API must describe it as an asserted time covered by the birth signature, not as externally witnessed time.

## Proposed Approach

Add a typed identity-evidence service and expose it through:

```http
GET /v1/identity
```

The response should be derived from authoritative material in the configured core root and existing `identity`, `provenance`, and `coreintegrity` machinery. It should not create a second identity store or trust calculation.

A versioned response should include at least:

- a stable instance identifier derived from the long-lived Ed25519 public key;
- the identity-key algorithm and SHA-256 fingerprint;
- the channel-CA fingerprint;
- the core birth commit object ID, including its Git object-hash algorithm;
- the birth time as an explicitly asserted/signed claim with an assurance classification that does not imply external timestamping;
- the current core HEAD object ID, including its hash algorithm;
- whether the core worktree is clean against HEAD;
- the admission result for the single birth and trust-file history;
- the current HEAD verification result;
- whether core is operator-anchored or self-signed, without exposing private key material or unnecessary operator-identifying metadata; and
- the time at which the live verification snapshot was produced.

Use explicit structural statuses such as `verified`, `failed`, and `unavailable`. Avoid a broad `trusted: true` or numeric trust score: a remote client has not independently established trust merely because the subject reports that its local checks passed.

The handler should remain a thin shell around package-level logic that is unit-testable without an HTTP server. Verification must use the request context and should return a bounded response.

Treat this as a first-class native contract:

[...]

Issue #1313, opened 2026-07-28
A loop cannot see its own sleep envelope until after it guesses


thane-developer:
A service loop chooses its next sleep with `set_next_sleep`, and nothing tells it what range it is allowed to choose within.

The tool description says a value is "clamped to the loop's configured `sleep_min` and `sleep_max` bounds" without naming them. The *result* carries `sleep_min`, `sleep_max`, `sleep_default`, and `clamped` — so the loop learns its envelope only after it has already asked for something, and since each iteration is a fresh conversation, that knowledge never survives to the decision it would inform.

The consequence was live until #1312. The metacognitive prompt told the loop to sleep "2–5m for active situations" against a 15m floor, so every such call was silently raised and the loop had no way to know its judgement had been overridden. Prompts had become the only place the envelope was written down, in prose, as a copy of a config value — and the copy went stale the moment the config moved. #1312 removed the numbers, which fixes the contradiction and leaves the loop reasoning about *which end* of a range it cannot see.

That is the right shape for a prompt and the wrong place for the fact. As the owner put it: knowing the bounds "communicates expectations and contributes to mental state for judgement". A loop deciding how long to sleep is making a judgement about its own cadence, and the envelope is a genuine input to it — not a validation detail to discover by being corrected.

## The seam

`set_next_sleep` is registered once on the global tool registry, so its advertisement cannot name per-loop bounds as it stands. But `Registry.WithRuntimeTools` copies the global set and re-`Register`s each runtime tool by name, so a loop-scoped tool shadows the global one — the mechanism `publish_output_*` already uses. Attaching a per-loop `set_next_sleep` whose description states that loop's actual envelope needs no new machinery.

Worth deciding along the way whether the envelope also belongs in the loop's per-iteration context rather than only in a tool schema. The tool description answers "what may I ask for"; a context line answers "what cadence am I on", which is the question the loop is actually reasoning about.

## Not just metacognition

Every timer-driven service loop has this. Ego and archivist ship with envelopes their prompts also described in prose until #1312, and any loop an operator authors through `thane_loop_create` or a `core/loops/` document has bounds its prompt has no way to reference.

Refs #1086

thane-developer:
Seam analysis, so the next attempt does not repeat it.

[...]

PR #1320, opened 2026-07-28
What Thane writes about itself lives in self, not core

thane-developer:
**Stacked on #1317** — merge that first; this branch contains its commits.

First implementation step of #1318.

## Why

#1316 shapes `core` into the instance's identity evidence: a birth commit, a signing identity, an admission result, a HEAD an operator can pin, and a posture claim about who has been entitled to change it since. That claim cannot hold while the agent writes to core every fifteen minutes.

The posture we want to promote by default is operator-only signers on `core`, with `config.yaml` shielded from agent editing. Agent-authored state needs the same integrity treatment core gets — signed history, admission, verification — under its own signer set. So it needs its own root, not a corner of core's.

## `self`

Derived from the workspace exactly as `core` is, for the same reason: the core service loops write it on every install, so a root an operator had to remember to declare would leave a fresh install with nowhere for those documents to land. `roots.self:` may still be named to set policy, and a path given there is ignored — core's existing rule, now applied to both through one code path instead of two copies of a special case.

Both derived roots are created at boot, so an operator can declare policy on a directory that exists.

## What moves, and what doesn't

`metacognitive` and `archivist` move here — their refs live in the shipped documents, so it is a ref change.

**`ego` does not.** Its document is read back through a core-scoped `provenance.Store` with revision history ([core_context.go:257](internal/runtime/agent/core_context.go:257)), not through a path, so moving it is a rewiring rather than a ref change. It gets its own PR.

## The container is renamed `self`

Its members are exactly the loops that write this root, so one name covers one idea: `parent: self` beside an output ref of `self:metacognitive.md` is two views of the same thing, where `cognition` and `self:` were two names for it. The constants stay separate — a loop container and a document root are different namespaces, and renaming one should not silently rename the other.

## Both prompts stop naming their own output ref

The "Declared Durable Outputs" block already states the ref verbatim, so the prose was a copy of a value — and it is the copy that just went stale, in this commit, exactly as the sleep durations did when the envelope moved in #1313. The prose keeps the part a prompt should carry: that the block names the document and the tool.

## Breaking

An existing install's loops now write `self:metacognitive.md` and `self:archivist.md`, so an upgrade starts them with empty documents unless the operator moves the files across. There is no automatic migration, deliberately: a boot-time copy would write into `core`, which is the direction this change exists to close.

## Test plan

- [x] `just ci` clean, no architecture drift, generated mirror and example config regenerated
- [x] `self` derives to `{workspace}/self`; a configured path for it is ignored, same as core
- [x] Parity tests hold across the ref and prose changes on both sides
- [x] Full `go test ./internal/... ./cmd/...` sweep green

## Follow-ups in #1318

- `ego.md` and its provenance-backed injection
- Generalizing the `coreintegrity` checks per-root, so `self` is held to core's standard rather than a weaker one
- `core` to `authoring: read_only` — last, since `Store.Write` gates the generated output tools on it

Refs #1318
Refs #1086

### DIFF
--- a/docs/understanding/document-roots.md
+++ b/docs/understanding/document-roots.md
@@ -414,13 +414,33 @@ of truth. The generated tools still write through document roots. That
 keeps path resolution, indexing, provenance, and root-level integrity
 policy in one subsystem.
 
-## Special Case: `core`
+## Special Case: the Derived Roots, `core` and `self`
 
-The `core:` root is reserved.
+Two roots are reserved. Both come from the workspace — `{workspace.path}/core`
+and `{workspace.path}/self` — and neither takes a path from `roots:`. You may
+still name either one there to set policy; a path given alongside policy is
+ignored.
 
-It always comes from `{workspace.path}/core` and is not configured
-manually in `paths:`. That is where Thane's always-on identity and core
-reference files live.
+They are two roots rather than one because they answer to different
+authorities, and that difference is the whole point of the split.
+
+`core` is what an operator declares Thane to be, and what constrains it:
+axioms, persona, mission, the runtime config, talents, and loop definitions.
+Nothing an agent writes belongs in it. That is what lets an install narrow
+`core`'s signers to the operator alone and set `authoring: read_only` on it —
+the posture we recommend, and the only way to shield `config.yaml` from
+agent editing. It is also what makes `core`'s HEAD worth reading as evidence:
+it moves when the operator moves it.
+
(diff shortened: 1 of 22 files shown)


thane-developer:
Superseded by #1321 — same commit, retargeted to `main`. GitHub auto-closed this one when #1317's branch was deleted on merge, and a closed PR whose base branch is gone cannot be reopened or retargeted.

PR #1396, opened 2026-08-14
fix: synthesize missing LM Studio tool call IDs

thane-developer:
## Summary

- Preserve LM Studio runner-supplied tool-call IDs.
- Synthesize a UUIDv7-backed `call_...` ID when structured or text-parsed tool calls arrive without one.
- Cover the complete Qwen tool round trip, including the assistant call and matching tool-result message on the next iteration.

## Why

The follow-on investigation from #1395 captured the actual failed production history. The assistant tool call had an empty `id`, and the following tool result had an empty `tool_call_id`. LM Studio accepted the first completion but rejected that history on the next iteration with `400 Invalid 'messages' in payload`.

The adapter previously preserved the empty runner value. Because both wire fields use `omitempty`, Thane then omitted both correlation fields from the follow-up payload. Assigning the fallback before the assistant message enters iteration history lets the existing engine carry one matching ID through the complete exchange.

LM Studio can also return a tool call as text for Thane's shared fallback parser to promote after structured decoding. The provider now normalizes IDs after that fallback too, covering this LM Studio-specific strictness without changing shared Ollama behavior.

## User impact

Qwen models running through LM Studio can continue after structured or text-parsed tool calls when the runner omits the call ID. Provider-assigned IDs remain unchanged, and parallel missing calls receive distinct IDs.

## Test plan

- [x] `just ci`
- [x] Structured and text-parsed missing-ID calls each complete a two-iteration assistant/tool-result round trip.
- [x] Multiple missing IDs are synthesized uniquely.
- [x] Existing streamed and non-streamed runner IDs are preserved.

Refs #1395

### DIFF
--- a/internal/model/fleet/providers/lmstudio.go
+++ b/internal/model/fleet/providers/lmstudio.go
@@ -366,7 +366,9 @@ func (c *LMStudioClient) handleStreaming(ctx context.Context, requestedModel str
 	result.Message.Role = normalizeLMStudioMessageRole(role)
 	result.Message.Content = contentBuilder.String()
 	result.Message.ToolCalls = toolCalls
-	applyTextToolFallback(result, validToolNames)
+	if err := applyTextToolFallback(result, validToolNames); err != nil {
+		return nil, err
+	}
 
 	c.logger.Debug("stream complete",
 		"model", result.Model,
@@ -407,7 +409,9 @@ func (c *LMStudioClient) chatResponseFromWire(wire *lmStudioChatResponse, validT
 	result.Message.Role = normalizeLMStudioMessageRole(wire.Choices[0].Message.Role)
 	result.Message.Content = lmStudioContentText(wire.Choices[0].Message.Content)
 	result.Message.ToolCalls = toolCalls
-	applyTextToolFallback(result, validToolNames)
+	if err := applyTextToolFallback(result, validToolNames); err != nil {
+		return nil, err
+	}
 	if strings.TrimSpace(result.Message.Content) == "" && len(result.Message.ToolCalls) == 0 {
 		return nil, fmt.Errorf("LM Studio returned an empty assistant completion for model %q", wire.Model)
 	}
--- a/internal/model/fleet/providers/lmstudio_test.go
+++ b/internal/model/fleet/providers/lmstudio_test.go
@@ -204,6 +204,33 @@ func TestToLMStudioMessages_EmptyContentIsEmittedNotOmitted(t *testing.T) {
 	}
 }
 
+func TestDecodeLMStudioToolCalls_SynthesizesUniqueMissingIDs(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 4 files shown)


thane-developer:
Confirming your root cause against the live runner, and flagging one path I think it leaves uncovered.

## The diagnosis holds

I chased this same 400 from the #1395 side and could not reproduce it, because I hand-wrote the repro JSON with explicit empty strings rather than letting the struct tags serialize it. That distinction turns out to be the entire bug. Against prod's centro, `qwen/qwen3-coder-next`, same four-message history each time:

| Payload shape | Result |
| --- | --- |
| `"id": ""` and `"tool_call_id": ""` present but empty | OK |
| both fields **omitted** — what `omitempty` actually produces | **400 Invalid 'messages' in payload** |
| synthesized `call_...` id on both, as this PR does | OK |

So it is not the empty value LM Studio objects to, it is the absent key, which is exactly what `lmStudioMessage.ToolCallID` and `lmStudioToolCallReq.ID` produce once the runner hands back a blank. Your fix is right and the round-trip test is the right place to hold it.

## The gap: text-parsed tool calls still arrive without an id

`ensureLMStudioToolCallID` covers the structured path, and `decodeLMStudioToolCallsFromSlice` routes through `decodeLMStudioToolCalls`, so streamed and non-streamed structured calls are both covered.

But `applyTextToolFallback` runs *after* the decode, at `lmstudio.go:411` and `lmstudio.go:452`, and fires only when the decode produced zero tool calls. The calls it builds come from `toolCallFromParts` in `internal/model/llm/toolcall_text.go:365`:

```go
return ToolCall{
	Function: struct {
		Name      string         `json:"name"`
		Arguments map[string]any `json:"arguments"`
	}{
		Name:      name,
		Arguments: args,
	},
}
```

No `ID`. So when LM Studio's parser emits the call as text rather than structured — which is the precise condition that fallback exists to handle — the call still enters iteration history with a blank id, `omitempty` still drops both correlation fields, and the next iteration still takes the 400. This PR touches only `lmstudio_wire.go`, so that route is untouched.

Worth knowing that `toolCallFromParts` is shared with the ollama adapter, so the gap is not LM Studio-specific; LM Studio is just strict enough to reject what the others tolerate.

## Not necessarily yours to fix

Synthesizing ids in the shared `llm` layer is a wider change than repairing the LM Studio decoder, and fixing the decoder here and the fallback separately is a perfectly reasonable split. Raising it mainly so the seam is known rather than assumed closed — as it stands the failure class survives on the text-fallback path.

Happy to take the follow-up if you would rather keep this PR tight.

_🤖 Posted by [Claude Code](https://claude.com/claude-code), from the #1395 / #1397 investigation_

thane-developer:
Addressed in `091757c5`. The LM Studio provider now normalizes IDs after the shared text-tool fallback, covering both streaming and non-streaming response paths without changing shared Ollama behavior. I added unit coverage and expanded the full round-trip regression to exercise both structured and text-parsed Qwen calls; `just ci` passes.

nugget (approved):

PR #1398, opened 2026-08-15
Tell a loop what it already follows, and that ~ is allowed

thane-developer:
Two model-facing gaps, both found by watching a live production loop rather than in review.

## A loop cannot see its own subscriptions

`thane_dev_watcher` called `forge_repo_follow` for a repository it was already subscribed to. It had no way to know it was: the injected forge block carries accounts and recent operations and says nothing about subscriptions. "Make sure I am subscribed" is a reasonable opening move on a wake, and the only way to actually check was a tool call the model had no reason to think it needed.

The block now carries the subscriptions themselves — repo, branch, checkout path, wake target, last synced SHA — narrowed by the account binding exactly as the account list is. Deliberately more compact than the `forge_repo_subscriptions` payload: this rides in every forge-tagged prompt, and its job is to answer "am I already watching this, and where is the checkout", not to replace the tool.

## Nothing told it `~` was allowed

While making that unnecessary call it passed `local_checkout=/Users/david/Thane/repos/thane-ai-agent` — a user that does not exist on that host. The schema said the path must be "absolute or working-directory-relative" and never mentioned that a leading `~` is expanded (support added in #1391 that nothing advertised), so a model needing an absolute path guessed one from the owner's name.

The description now names the `~` form and says why to prefer it: an absolute path under the wrong user is not an error, it is a real location nobody intended.

## Note

The duplicate-admission gate from #1394 is what kept this harmless. Because the store's checks now run before the clone, the call failed on "already exists" rather than attempting to create a tree under a nonexistent user. The previous ordering would have tried the mkdir first.

## Test plan

- [x] `just ci` green locally
- [x] A bound loop sees its own subscriptions in context, including checkout path and last synced SHA
- [x] Subscriptions are narrowed by the account binding — another account's are not leaked
- [x] An unbound reader sees all of them

### DIFF
--- a/docs/reference/tools.md
+++ b/docs/reference/tools.md
@@ -378,7 +378,11 @@ Repository subscriptions require `wake_loop` so event handling is owned by
 an existing loop, usually one created with `thane_loop_create`
 (`operation: service`) for a specific managed document. Pass
 `local_checkout` when the loop should read a local mirror checkout; the
-poller syncs that checkout before delivering repository events. The path
+poller syncs that checkout before delivering repository events. The
+path may be absolute, working-directory-relative, or `~`-prefixed; a
+leading `~` expands against the account Thane runs as, which is the
+form to prefer — an agent that guesses at a home directory can write a
+real path under the wrong user. The path
 must be empty or an existing Thane-owned mirror checkout; non-empty
 directories and unmarked git checkouts are refused. Unfollowing leaves
 the checkout on disk. The follow performs the initial clone itself, so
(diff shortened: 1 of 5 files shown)


thane-developer:
Superseded by #1399, per discussion.

The `~` teaching in `local_checkout` is the part that clearly goes away: if a repository subscription exposes itself as a named root, there is no path argument left to document, and teaching models to write `~/...` is solving a problem #1399 removes. The subscription-context injection is subsumed too — a named root surfaces the same knowledge in a better shape.

Both ideas are carried into #1399 rather than lost: that a loop must be able to see what it already follows without a tool call, and that a model needing an absolute path will invent one from whatever name is nearby. Production has now produced the `/Users/david/` guess twice — once through `forge_repo_follow` and once through a file tool (`path escapes allowed directories: /Users/david/Thane/repos/thane-ai-agent`), so the failure is a property of handing models paths, not of one tool's schema.

PR #1402, opened 2026-08-15
feat(forge): expose subscriptions as named roots

### LINKED ISSUE #1399: A repository subscription should expose itself as a named root, not a filesystem path
A repository subscription already knows everything a model needs in order to read the code it watches: which remote it tracks, which branch, where the mirror lives on disk, and which commit it was last synced to. It exposes almost none of it. What reaches the model is a filesystem path, and a path is the one part of that knowledge the model is least equipped to hold — it cannot verify the path, cannot tell a stale one from a current one, and cannot construct a new one without guessing at a machine it has never seen. Production supplied the proof within hours of the checkout feature landing: a watcher loop, asked to follow a repository whose checkout it already had, sent `local_checkout=/Users/david/Thane/repos/thane-ai-agent` — a home directory belonging to the owner's first name rather than to the account Thane runs as, on a host where that user does not exist. The model was not being careless. It needed an absolute path, nothing in its context contained one, and the tool schema told it a path was what to supply.

The fix that generalizes is not better path documentation. It is to stop handing models paths at all. A repository subscription should expose itself as a **named root** — the same primitive document corpora already use — so that the model addresses code the way it addresses everything else in the system, by name, and the filesystem stays a detail that Go owns. This is the [Offload cognition to Go](docs/model-facing-context.md) principle applied to the one surface that has been quietly exempt from it: if Go can derive, normalize, or disambiguate something, it should, and where a checkout physically sits is exactly that kind of something.

## The primitive mostly exists

A root entry today carries a path, an indexing flag, an authoring policy, git configuration, seed signers, and a context policy. Declaring a checkout as `indexing: false, authoring: read_only` produces something that already behaves like a code root: the prefix resolver handles `thanecode:internal/app/new.go`, and `file_read`, `file_grep`, and `file_tree` all accept it. We have been doing this by hand. The gap is not a missing primitive but two missing pieces around it — nobody registers the root automatically, and the tools reachable through it are shaped for prose rather than source.

This is why the proposal should not be a parallel `reporoots:` block sharing most of the same codepaths. Two structures that share most of their behavior reliably become two structures that disagree about the rest; this repository has already paid that bill more than once, most recently when a binding cascade reimplemented in `internal/tools` began to diverge from the real one in `internal/runtime/loop`, and the bug being fixed at the time was a definition view and a live view disagreeing about what an ancestor chain meant. A repository root should be a *policy shape* on the root primitive, not a second primitive.

## The subscription registers the root

Following a repository with a checkout should register a named root as part of the follow, not as a separate act of configuration an operator remembers afterward. The name is the model-facing handle; the path becomes something Thane chooses under its own workspace and never asks anyone to type. That closes the `/Users/david` class of failure completely rather than documenting around it — there is no path argument left to get wrong — and it removes the drift where a subscription's `checkout_path` and a hand-declared root's `path` slowly cease to agree.

It also inverts a piece of the current contract in a healthy direction. Today `local_checkout` asks the caller for a filesystem location and, per the guidance added alongside the initial-clone fix, expects someone to declare a read-only root afterward so consumers can stop passing paths around. That is two steps and a convention. One step and a name is better, and the name is the thing the model was going to need anyway.

## Tools a coding model already knows

A named root is only useful if what it exposes matches what coding models reach for. This is [Model families get adapters, not one universal prompt](docs/model-facing-context.md) read forward rather than backward: the point is not to invent an elegant novel surface and hope a broad registry adapts to it, but to offer the shapes that local and cloud coding models alike have already been trained on, so that a qwen-coder variant and a frontier cloud model both find the affordance obvious.

Concretely, the surface is smaller than it sounds, because most of it exists and needs sharpening rather than inventing:

**Content search with a filename filter.** `file_grep` today takes a pattern, a path, a depth, and a case flag, and has no include-glob. Searching a Go repository therefore returns matches from every file type and spends model attention on filtering by eye — precisely the cognition the conventions say to move into Go. A `file_pattern` argument is a small change and the single highest-value one on this list.

**Read-only git history scoped to the root.** `log`, `diff`, `show`, and `blame`, confined to a declared repository root, answering the question a code watcher actually asks on every wake: what changed since last time. The subscription already records `last_synced_sha`, and that value is now injected into the loop's context, so the range is available without the model computing anything. This is also the capability that makes shell access unnecessary for this use case. On the bare host `exec` runs `sh -c` as the service account with no enclosure, and prod configures neither denied patterns nor allowed prefixes, so granting it to a read-only watcher would let that watcher read the write token out of `config.yaml` and defeat its own forge-account binding. A scoped git tool reaches the same goal through capability rather than isolation, and stays correct if a container arrives later.

**Structure before content.** A file tree and a symbol or definition lookup are the natural third piece, but they are worth deferring until grep-with-globs and git history have been used in anger, because they are where the design could grow speculative.

## Binding a loop to a repository

Once a subscription is a named root, a loop can be scoped to it the same way loops are already scoped to a forge account: `bindings: {repo_root: thanecode}`. This is the mechanism from #1386 reaching its second key, and it is a better second key than the candidates that arc left on the table, because the need is real rather than illustrative.

The shape fits without modification. Binding keys are named for what the value names rather than for the subsystem, so `repo_root` sits beside `forge_account` consistently. Resolution is singular — a watcher watches one repository — which is the arity the current `map[string]string` already expresses, so this key needs none of the deferred work around set-valued bindings. Enforcement lands where the root is resolved, the same way forge enforcement lands where the account is resolved: an omitted root argument resolves to the binding rather than to some default, and naming a different root is refused with an error that says which root the loop is bound to. Everything the earlier arc built downstream comes along unchanged — the ancestors-win cascade so a container can scope a whole family of loops, the launch guard that stops a caller choosing a differently-bound parent, the delegate propagation so a bound loop cannot delegate its way out, and the context narrowing so the injected block advertises only the root the tools will actually serve.

The two bindings compose into something neither achieves alone. A watcher bound to `github-readonly` and to `thanecode` is scoped in both dimensions that matter: which credential it acts with, and which corpus it can read. Neither binding is enforced by the other's mechanism, and both are declared in the same place in the same grammar, which is the property that makes a loop definition legible as a statement of what that loop is allowed to be.

It is worth being explicit about what this does not do, because the forge arc taught the lesson expensively. A repository-root binding restricts which corpus Thane's own tools will serve. It is not a filesystem sandbox. A loop that also holds `exec` reaches any path the service account can read regardless of what it is bound to, which is the same reason granting shell to a forge-bound watcher would defeat its account binding. Binding scopes tools; enclosure scopes processes; they are different guarantees and the difference should stay legible in whatever documents this.

## The part that is genuinely different

Everything above extends the document-root primitive comfortably. One thing does not, and it deserves the design attention rather than the tooling: **trust means something different for code than for documents.**

A document root's contents are trusted because they are signed, and the verification machinery assumes exactly that — content is attested by a signature over history, and a worktree with uncommitted changes is refused because unattested bytes must not reach a prompt. A repository mirror has none of that shape. Its contents are trusted because they match a remote, it is reset hard on every poll so local modifications are discarded by design, and the meaningful assertion is not "someone signed this" but "this is `<sha>` of `<remote>`, fetched at `<time>`."

Reusing `verify_signatures` semantics on a repository root would therefore be quietly wrong in both directions: it would refuse a healthy mirror for having no local signing story, and it would fail to assert the thing that actually matters, which is provenance of the remote and the identity of the commit. A repository root should assert *what it is a copy of*, and the model-facing context should carry that — remote, branch, commit, and how stale the copy is, rendered as a delta the way other recency-sensitive context already is rather than as a raw timestamp.

This is the question that decides whether the feature is an extension or something closer to a sibling. Everything else argues for extension.

## What this is worth

The immediate beneficiary is the development watcher loop that prompted it, which currently reads a checkout through absolute paths, greps a Go repository without being able to say `*.go`, and had to be told twice not to invent a subscription it already had. The broader case is that Thane is increasingly asked to reason about code — its own, and the projects it watches — and every coding model that arrives will want the same four or five affordances. Building them once against a named root, in the shapes those models already know, is what makes the next model useful on arrival rather than after a round of prompt archaeology.

## Where this touches the code

Recorded so a session picking this up cold does not have to rediscover it.

The root primitive is `config.RootEntry` in `internal/platform/config/config.go` — path, indexing, authoring, git, seed signers, context policy — and prefix resolution runs through `internal/platform/paths`. `FileTools.resolvePath` in `internal/tools/file_tools.go` is where a prefixed path becomes an absolute one and where root membership is decided; it also treats anything under `workspace.path` as writable, which is why a checkout living inside the workspace is writable today unless a root declares otherwise. `file_grep` is registered in `internal/tools/tools.go` around the 586 mark and is the tool that wants the filename filter.

On the forge side, `forge_repo_follow` lives in `internal/integrations/forge/tools_subscriptions.go`; it now performs the initial clone through the same syncer the poller uses and checks store admission before touching disk. `checkout.OpenMirror` and `Mirror.Sync` in `internal/platform/checkout` own the mirror lifecycle, and `Sync` is what resets the worktree hard. The subscription record carries `CheckoutPath`, `CheckoutRemoteURL`, `Branch`, and `LastSyncedSHA`, and those now reach the model through the forge context provider in `internal/integrations/forge/context.go`.

The binding machinery is `internal/runtime/loop/bindings.go` — registered keys, the ancestors-win merge, and the context transport — with enforcement modelled at `Service.ResolveAccount` in `internal/integrations/forge/service.go`, the launch guard in `internal/tools/loop_launch_binding_guard.go`, and delegate propagation in `internal/runtime/delegate/delegate.go`. Adding `repo_root` means registering the key, honoring it where a root is resolved, and narrowing the injected context to match; the surrounding cascade, guard, and propagation need nothing new.

Signature verification for document roots is `internal/platform/provenance/verify.go` and `internal/state/documents/verification.go`, which is the code the trust section above says should *not* simply be reused. A repository root asserts what it is a copy of, not who signed it.

thane-developer:
## Summary

- replace model-supplied checkout paths with automatically registered, read-only repository roots whose physical paths are derived under `workspace.path`
- add filename-filtered `file_grep` plus bounded, root-scoped `repo_git_log`, `repo_git_diff`, `repo_git_show`, and `repo_git_blame` tools
- add the `repo_root` loop binding, enforce it at file/content/history/subscription lookups, narrow forge context to the bound root, and migrate persisted checkout subscriptions to deterministic root names
- keep repository mirrors outside document indexing and signature verification while exposing their remote, branch, commit, and exact sync age as provenance

## Why

Repository subscriptions already know where their mirrors live, but the old contract asked the model to invent and retain a host filesystem path. That made source access machine-specific and easy to misconfigure. A named root lets Go own path derivation and gives coding models the source-search and history shapes they already know without granting shell access.

## User impact

Callers use `repo_root: thanecode` instead of `local_checkout`. A successful follow clones immediately, registers `thanecode:` as read-only, returns no host path, and keeps the root current before subscription wakes. Loops can bind to `repo_root: thanecode`; relative file paths and omitted repository-history roots then default to that root, while sibling roots and path escapes are refused. Existing persisted checkout subscriptions are migrated on startup.

## Test plan

- [x] `just ci`
- [x] real local Git repository coverage for log, diff, show, blame, revision option injection, path escape, output bounding, and checkout-path redaction
- [x] named-root follow, initial clone, polling freshness, event metadata, unfollow, restart migration, collision, and rollback coverage
- [x] root-binding coverage across file tools, universal content resolution, subscription listing/removal, context narrowing, and managed-document exclusion

Closes #1399


internal/tools/file_tools.go
```diff
@@ -146,20 +147,38 @@ func (ft *FileTools) verifyMutationPath(ctx context.Context, absPath, consumer s
...
-	// Resolve registered prefixes (kb:, scratchpad:, etc.)
+	boundName := looppkg.BindingFromContext(ctx, looppkg.BindingRepositoryRoot)
+	var boundRoot paths.Root
+	if boundName != "" {
+		var ok bool
+		boundRoot, ok = ft.resolver.Root(boundName)
+		if !ok || boundRoot.Kind != paths.RootKindRepository {
+			return "", false, fmt.Errorf("this loop is bound to repository root %q, but that root is not registered at this site", boundName)
+		}
+	}
```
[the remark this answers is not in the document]
thane-developer: `paths.Resolver` query methods are deliberately nil-safe, so neither `Root` nor `RootForPath` dereferences receiver state when the resolver is absent. The bound read already returns an unavailable-root error, and unconfigured grep/display formatting already falls back to workspace-relative paths. Commit `fea7c13e` adds targeted regression coverage for the bound no-resolver path; `just ci` passes.

internal/integrations/forge/repository_roots.go
```diff
@@ -0,0 +1,212 @@
...
+// subscriptions that predate repo_root receive a deterministic handle and are
+// rewritten once so subsequent boots keep the same model-facing name.
+func (s *Service) registerPersistedRepositoryRoots() (err error) {
+	if s == nil || s.subscriptions == nil {
+		return nil
+	}
+	type registration struct {
+		name  string
+		owner string
+	}
```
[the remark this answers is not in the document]
thane-developer: `Root` is nil-safe and `registerRepositoryRoot` explicitly rejects a nil resolver, so startup returns a configuration error rather than panicking. Commit `fea7c13e` adds persisted-subscription coverage for that exact no-resolver restore path; `just ci` passes.

internal/integrations/forge/repository_roots.go
```diff
@@ -0,0 +1,212 @@
...
+			name = s.legacyRepositoryRootName(sub)
+			sub.RepositoryRoot = name
+			migrated = true
+		}
+		name, err = repositoryRootName(name)
+		if err != nil {
+			return fmt.Errorf("restore repository root for subscription %q: %w", sub.ID, err)
+		}
+		_, existed := s.rootResolver.Root(name)
+		if err := s.registerRepositoryRoot(name, sub.CheckoutPath, sub.ID); err != nil {
```
thane-developer: **A single unregisterable root stops the whole service from booting.**

`registerPersistedRepositoryRoots` is called from `NewService` (service.go:85) and its error is returned; `initChannels` wraps it as `create forge service` (new_channels.go:194) and `app.New` returns it (new.go:115). So any failure in this loop aborts startup entirely.

Two reachable triggers:

1. The operator comments out `workspace.path`. `cfg.Paths` is then empty, `paths.New` returns nil (new_agent.go:48-52), `s.rootResolver` is nil, and `registerRepositoryRoot` errors with `named-root resolver is unavailable` for every existing subscription that carries a `CheckoutPath`. Before this change that configuration started fine — the checkout was simply unreachable from file tools.
2. A persisted `repo_root` name later collides with a configured `roots:` entry of the same name. `Resolver.Register` refuses because the existing `Root` differs, and the collision avoidance in `legacyRepositoryRootName` only runs for subscriptions that have no `RepositoryRoot` yet, so a name already written by an earlier `forge_repo_follow` gets no second chance.

In both cases Thane refuses to boot and recovery means hand-editing config or opstate. Logging and skipping the individual subscription would keep the blast radius at one root instead of the process.

internal/integrations/forge/tools_subscriptions.go
```diff
@@ -105,24 +107,42 @@ func (t *Tools) HandleRepoFollow(ctx context.Context, args map[string]any) (stri
 	if branch == "" {
 		branch = meta.DefaultBranch
 	}
-	localCheckout := strings.TrimSpace(stringArg(args, "local_checkout"))
+	if legacyPath := strings.TrimSpace(stringArg(args, "local_checkout")); legacyPath != "" {
+		return "", fmt.Errorf("local_checkout is no longer accepted because models cannot safely choose host paths; set repo_root to a named handle and Thane will choose the checkout location")
+	}
+	repositoryRoot := strings.TrimSpace(stringArg(args, "repo_root"))
+	if bound := boundRepositoryRoot(ctx); bound != "" {
```
thane-developer, line 120: **The binding defaults an argument whose absence used to mean "do not create a checkout at all".**

Line 118 sets `repositoryRoot = bound` whether or not the caller passed `repo_root`, so the block at line 121 always runs for a bound loop: it demands a branch and a clone URL, derives `<workspace>/repos/<bound>`, and then `registerRepositoryRoot` at line 225 fails because that name is already owned by the subscription the binding points at.

Concretely: a loop bound to `repo_root: thanecode` calls `forge_repo_follow` on `nugget/other-repo` to track releases only, passing no `repo_root`. It gets `root "thanecode" is already registered`, which names neither the repository it was trying to follow nor the fact that it never asked for a checkout. A bound loop therefore cannot do a release-only follow, and cannot follow a second repository at all.

"Omitted defaults to the binding" is the right rule for tools that *read* a root. It is the wrong rule for a tool that *creates* one — here the binding should constrain which root may be named, not synthesize a request the caller did not make.

internal/tools/file_tools.go
```diff
@@ -192,13 +211,32 @@ func (ft *FileTools) resolvePath(path string) (string, bool, error) {
...
 	}
 
+	if boundName != "" && (!paths.ContainsPath(boundRoot.Path, absPath) || !paths.ContainsPath(boundRoot.Path, realPath)) {
+		return "", false, fmt.Errorf("path is outside repository root %q; this loop's repo_root binding restricts file tools to root %q", boundRoot.Name, boundRoot.Name)
+	}
+	if matched && (!paths.ContainsPath(matchedRoot.Path, absPath) || !paths.ContainsPath(matchedRoot.Path, realPath)) {
+		return "", false, fmt.Errorf("path escapes named root %q through a symlink", matchedRoot.Name)
+	}
+
+	readOnly := matched && matchedRoot.ReadOnly
```
thane-developer, line 222: **Read-only roots now short-circuit the workspace sandbox, and migrated roots were model-chosen paths.**

The early `return realPath, true, nil` at line 229 fires before the workspace check at line 232 and before the `read_only_dirs` loop, so anything `RootForPath` matches is readable regardless of workspace policy.

That is defensible for newly derived roots, which Go places under `workspace.path`. It is not defensible for migrated ones. Pre-PR, `forge_repo_follow` accepted `local_checkout` as an arbitrary model-supplied absolute path with no containment check — `OpenMirror` only tested empty-or-Thane-owned — and `resolvePath` would still refuse to read it unless it happened to land under `workspace.path` or `read_only_dirs`. `registerPersistedRepositoryRoots` now registers that same legacy `CheckoutPath` verbatim, so on upgrade a checkout the model parked at, say, `~/elsewhere/tree` becomes readable through `file_read` and `file_grep` where it was refused before.

The comment at line 225 justifies the bypass with "the resolver is trusted Go configuration, not a path the model supplied" — which is precisely backwards for the migrated checkouts the same sentence invokes. Either validate migrated `CheckoutPath`s against `workspace.path` at registration time (and refuse or relocate the ones that fail), or keep the sandbox check on this path.

internal/integrations/forge/context.go
```diff
@@ -112,21 +130,62 @@ func (p *ContextProvider) buildContext(bound string) (string, error) {
 			", which is not configured at this site. No forge operation can succeed until the operator restores the account or changes the binding."
 	}
 
+	if p.service.subscriptions != nil {
+		subs, err := p.service.subscriptions.List()
+		if err != nil {
+			return "", fmt.Errorf("list repository roots for forge context: %w", err)
```
thane-developer: **A store read failure here discards the forge context block that was already built.**

Returning an error abandons the account views assembled twenty lines earlier, so a transient opstate failure leaves the loop's prompt with no forge block at all — the model reads that as "no forge configured at this site" and behaves accordingly, which is a worse outcome than not knowing about the repository roots.

The roots section is supplementary to the accounts. A failure to enumerate it should degrade to omitting the field, ideally with a short note in the JSON the way `binding_error` already does, rather than taking the accounts down with it.

Separately: this puts a full `subscriptions.List()` — one index read plus one read per subscription — on every prompt assembly for every forge-tagged loop. Worth caching against `LastChecked` rather than paying it per turn.

internal/tools/file_tools.go
```diff
@@ -146,20 +147,38 @@ func (ft *FileTools) verifyMutationPath(ctx context.Context, absPath, consumer s
...
+	var matchedRoot paths.Root
+	var matched bool
 	if ft.resolver != nil {
-		resolved, err := ft.resolver.Resolve(path)
-		if err != nil {
-			return "", false, fmt.Errorf("resolve prefix: %w", err)
+		resolved, root, ok := ft.resolver.ResolveRoot(path)
+		if ok {
+			if boundName != "" && root.Name != boundRoot.Name {
+				return "", false, fmt.Errorf("repository root %q is not available here: this loop is bound to root %q; retry with %s: or omit the root prefix", root.Name, boundRoot.Name, boundRoot.Name)
```
thane-developer: **This refusal calls every root a repository root, including the ones that aren't.**

`root` here is whatever `ResolveRoot` matched, which includes `kb:`, `core:`, `self:`, and `scratchpad:`. A loop bound to `repo_root: thanecode` calling `file_read("scratchpad:notes.md")` is told:

> repository root "scratchpad" is not available here: this loop is bound to root "thanecode"; retry with thanecode: or omit the root prefix

Scratchpad is not a repository root, and the suggested retry cannot find the file. Per `docs/model-facing-tools.md` this is the shape of refusal that costs the next turn a guess.

Two related gaps worth deciding on explicitly, since the behaviour itself looks intentional:

- A `repo_root` binding silently revokes *all* document-root file access for that loop. `TestFileToolsRepositoryRootBindingDefaultsAndRefusesEscape` only exercises a sibling repository root (`othercode:`), so nothing pins this.
- `docs/reference/loop-definitions.md` describes the refusal as covering "sibling roots", which reads naturally as sibling *repository* roots. If a bound loop really does lose `core:` and `scratchpad:`, that deserves a sentence of its own.

internal/tools/file_tools.go
```diff
@@ -626,6 +660,23 @@ func (ft *FileTools) Grep(ctx context.Context, dir, pattern string, maxDepth int
 	return strings.Join(results, "\n") + warning, nil
 }
 
+func (ft *FileTools) displayPath(path string) string {
```
thane-developer: **`EvalSymlinks` per root per matched file, inside the grep and search walks.**

`displayPath` calls `RootForPath`, which iterates every registered root calling `ContainsPath` → `RelativePath` → `resolveSymlinksBestEffort`. `EvalSymlinks` lstats every component of both sides, so a ten-component path against six roots (core, self, kb, scratchpad, plus repository roots) is on the order of 120 syscalls per matching file — repeated for every file in the result set.

This replaced a single `filepath.Rel` against a `workspaceAbs` that was computed once outside the walk, so it is a real regression on the hot path of exactly the tool this PR is trying to make good at source search. Resolving the root list once before the walk, or memoizing `resolveSymlinksBestEffort` per directory, restores the old profile without giving up the prefixed display paths.

thane-developer (commented):
Read the full diff plus the surrounding call sites: `app.New` init ordering, `SubscriptionStore` admission/validation, `checkout.Mirror`, and the pre-PR `local_checkout` handling.

The design holds up. Letting Go own path derivation is clearly right, and replacing `strings.HasPrefix` with `RelativePath`/`ContainsPath` fixes a real sibling-directory boundary bug on the way past. Several things I chased turned out to be fine and are worth recording so nobody re-chases them: the rollback path in `HandleRepoFollow` is safe because `CheckAdmission` rejects a duplicate subscription ID before `registerRepositoryRoot` runs, which makes `Register`'s idempotent-nil branch unreachable there; `cappedBuffer`'s truncation accounting is correct at the boundary; `resolveRepositoryRevision` is properly hardened with `--end-of-options`; the blame porcelain parser cannot mistake a `previous` line for a header; and `legacyRepositoryRootName`'s collision loop converges and cannot emit a name that `repositoryRootName` would later reject.

The six comments below are about blast radius at the edges rather than the core contract. Two of them (the boot abort and the sandbox widening) are the ones I would want resolved before this merges.

internal/integrations/forge/repository_roots.go
thane-developer: Fixed in `b75d9f42`. Persisted-root restore now isolates validation, resolver, collision, and migration-write failures per subscription: it logs and skips only the unavailable root while continuing with the rest. Legacy paths outside `workspace.path` are also skipped. Regression coverage includes no resolver, a configured-root collision, an outside-workspace legacy checkout, and a good sibling root continuing to register. `just ci` passes.

internal/tools/file_tools.go
thane-developer: Fixed in `b75d9f42` with both defenses. Persisted checkout roots are registered only when their symlink-aware path remains under `workspace.path`, and read-only root policy no longer short-circuits file-tool admission through `workspace.path` / `read_only_dirs`. Tests cover an outside legacy checkout being skipped and an externally registered read-only root remaining unreadable unless explicitly listed in `read_only_dirs`. `just ci` passes.

internal/integrations/forge/context.go
thane-developer: Fixed in `b75d9f42`. Subscription enumeration failures now preserve the forge account block and add a structured `repository_roots_error`; a bound root does not get a misleading missing-binding error when enumeration itself failed. I removed the per-turn N+1 rather than adding a stale cache: `SubscriptionStore.List` now reads the namespace once and decodes the indexed rows from that snapshot. Regression coverage closes the database and verifies both account views survive. `just ci` passes.

internal/integrations/forge/tools_subscriptions.go
thane-developer: Fixed in `b75d9f42`. A `repo_root` binding now constrains an explicitly supplied handle but no longer fills an omitted creation argument. Omission stays event-only; a mismatched explicit root is refused with guidance that distinguishes the two cases. The tool schema, reference docs, and a bound release-only follow regression test are updated. `just ci` passes.

internal/tools/file_tools.go
thane-developer: Fixed in `b75d9f42`. Search and grep now snapshot the root roster and canonicalize each root plus the workspace once per operation. Match formatting uses only lexical `filepath.Rel` checks against that snapshot, ordered most-specific first, so there is no per-root `EvalSymlinks` work per matched file. Existing named-root search/grep coverage remains green under `just ci`.

internal/tools/file_tools.go
thane-developer: Fixed in `b75d9f42`. The refusal now calls the requested item a named root and explicitly says that `repo_root` restricts all file tools to the bound repository; it no longer suggests retrying a document path under the repository prefix. Tests pin `scratchpad:` refusal and ensure it is not mislabeled, and the loop/tool references now state that document roots are unavailable to a repository-bound file caller. The boundary itself remains intentionally unchanged. `just ci` passes.

internal/integrations/forge/tools_subscriptions.go
```diff
@@ -199,18 +220,38 @@ func (t *Tools) HandleRepoFollow(ctx context.Context, args map[string]any) (stri
 	if err := t.subscriptions.CheckAdmission(sub); err != nil {
 		return "", err
 	}
+	rootRegistered := false
+	if sub.RepositoryRoot != "" {
+		if err := t.service.registerRepositoryRoot(sub.RepositoryRoot, sub.CheckoutPath, sub.ID); err != nil {
+			return "", err
+		}
+		rootRegistered = true
```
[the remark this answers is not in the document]
thane-developer: Fixed in `da0c5069`. Dynamic registration now atomically reports whether the current call inserted the root. Follow rollback acts only on an inserted root and also preserves it when an identical concurrent follow has become the durable subscription, covering the inverse interleaving where the physical inserter loses the final `Add` race. The regression test exercises that exact winner/loser sequence. `just ci` passes.

internal/integrations/forge/repository_roots.go
```diff
@@ -0,0 +1,212 @@
...
+		return "", fmt.Errorf("repo_root is required")
+	}
+	if len(name) > 64 {
+		return "", fmt.Errorf("repo_root %q is too long; use at most 64 characters", name)
+	}
+	for i, r := range name {
+		if unicode.IsLetter(r) || unicode.IsDigit(r) || (i > 0 && (r == '.' || r == '_' || r == '-')) {
+			continue
+		}
+		return "", fmt.Errorf("repo_root %q is invalid; use letters, digits, '.', '_', or '-', starting with a letter or digit", name)
```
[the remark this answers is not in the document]
thane-developer: Fixed in `da0c5069`. Repository handles now canonicalize to lowercase and accept only portable lowercase ASCII; follow arguments and persisted roots are normalized before registry use, so case variants cannot produce distinct handles for one case-insensitive checkout. The model-facing schema, operator docs, and restore/follow tests cover uppercase input and persistence migration. `just ci` passes.

internal/platform/paths/resolver.go
```diff
@@ -86,14 +207,114 @@ func (r *Resolver) Prefixes() []string {
...
+	defer r.mu.RUnlock()
+	var best Root
+	bestLen := -1
+	for _, root := range r.roots {
+		if !ContainsPath(root.Path, path) || len(root.Path) <= bestLen {
+			continue
+		}
+		best = root
+		bestLen = len(root.Path)
+	}
```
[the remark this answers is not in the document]
thane-developer: Fixed in `da0c5069`. `RootForPath` now canonicalizes the candidate and every root once, performs lexical containment on those normalized absolute paths, and compares normalized lengths (with a deterministic name tie-break). A regression test uses a relative nested root inside an absolute outer root and verifies the nested root wins. `just ci` passes.

nugget (approved):

Issue #991, opened 2026-05-27
Wire dossiers (kb:dossiers/...) into model-visible retrieval


thane-developer:
## Background

Architecture context: #986 (umbrella). Follow-up to #985 (curator loop) raised during review by @thane-developer.

The curator (#985) authors dossiers as managed documents under \`kb:dossiers/\`. They live in the documents store and pick up that store's indexing automatically. But the multi-surface \`archive_search\` envelope from #983 (\`MemorySearcher\`) currently searches three surfaces:

- raw messages (messages_fts)
- session summaries (sessions_fts)
- working memory (working_memory_fts)

It does **not** search KB documents — including the dossier namespace. So a curator can produce excellent per-subject synthesis that the interactive agent rarely sees during semantic search.

## Fix shape

Either:

1. **Extend \`MemorySearcher\` to surface documents.** Add a fourth slice to \`SearchBundle\` — \`Documents []DocumentMatch\` — populated from \`documents.Tools.Search\` (already exists) and rendered alongside the other surfaces in \`FormatMultiKindResults\`. This makes the curator's output visible exactly where the interactive agent looks for memory.

2. **Subject-keyed dossier lookup at wake time.** A new context provider that, when the wake's subjects include any handle that has a matching dossier file at \`kb:dossiers/<subject>.md\`, injects the dossier directly into the system prompt. This is more targeted (only fires for subjects the wake actually carries) but doesn't help the model-initiated \`archive_search\` path.

(1) is the natural fit with the unified retrieval architecture in #986 and #983. (2) is complementary and could also land — they're not mutually exclusive.

## Acceptance criteria (for option 1)

- [ ] \`SearchBundle\` carries a \`Documents []DocumentMatch\` slice.
- [ ] \`MemorySearch.Search\` populates documents from the documents store, scoped to roots that are sensible memory targets (at minimum \`kb:\`; arguably all roots).
- [ ] \`FormatMultiKindResults\` emits \`documents\` alongside the other three surfaces.
- [ ] \`archive_search\` tool description updated to name the documents surface.
- [ ] \`talents/archive.md\` updated to describe documents as one of the four searched surfaces.
- [ ] Per-surface conversation_id scoping rules: documents are typically NOT conversation-scoped, so honor only when the document carries a conv_id marker.
- [ ] Byte-cap fitter's two-stage shrink extended to documents.

## Dependencies

[...]

PR #1410, opened 2026-08-17
feat(context): add context advertisements

thane-developer:
## Summary

- Add a validated context-advertisement contract with request-relative match evidence and bounded `signal`, `context`, and `detail` projection roles.
- Add a deterministic final discriminator that ranks, deduplicates, filters, and limits offers before lazy materialization; legacy eager providers remain compatible while they migrate.
- Make `signal` the native compact facet across runtime code, tools, APIs, loop definitions, and model-facing guidance; it remains mandatory for every faceted output, and system self-assessment is the first end-to-end advertiser.

## Why

Archivist dossiers need a generic way to compete for model attention without making a dossier root magical or eagerly injecting full documents. This establishes the producer/consumer boundary for #991: subsystems own domain matching and materialization, while Go owns cross-source policy, stable ordering, and hard context limits.

Using `signal` directly in the faceted-document contract keeps that vocabulary aligned with the consumer role. It avoids a compatibility layer, a synthetic dossier status, or special-case projection rules.

## User impact

This PR does not yet advertise or retrieve dossiers. It lays the shared path they can use next. The existing `status_line` facet is renamed mechanically to `signal` everywhere current production code consumes or publishes it, including metacognitive self-assessment and the loop-output API. There is deliberately no alias or fallback for the old name.

## Test plan

- [x] `just ci`
- [x] Contract validation covers invalid identity, match evidence, projection metadata, estimates, and the mandatory `signal` facet.
- [x] Discriminator tests cover stable evidence ordering, signal projection choice, context fallback, deduplication, count/byte limits, and exclusion of automatic detail.
- [x] Assembly test proves a selected signal survives and precedes a full legacy context bucket.
- [x] System self-assessment test proves advertising performs no document read and materialization performs exactly one.
- [x] Facet, document, generated-tool, and HTTP tests exercise `signal` as the native published key and section.

Refs #991


thane-developer:
This PR's foundation is being landed via #1433, which cherry-picks `f1bc157a` (authorship preserved) onto current main — the contract, discriminator, wiring, and the self-assessment advertiser, exactly as authored.

The second commit (`dc93256c`, `status_line`→`signal`) is deliberately not carried forward. The design review that revived this work found the rename's "one vocabulary" premise doesn't survive contact with the first commit itself: `f1bc157a` maps both `status_line` *and* `teaser` to role `signal`, so post-rename "signal" would name one facet and simultaneously a role class two facets fill — a collision that erases the signal-vs-teaser distinction for the authoring model. Since roles are Go-internal, keeping `status_line` costs the reading model nothing, and dropping the rename deletes the whole production migration (heading rewrites in the schedule root and metacog docs, opstate facet arrays, the quiet-until-republish self-assessment window, `/v1` consumer breaks, talents backport).

Full charter for what builds on this: #1431 (advertisement rail) and #1432 (mechanical calendar context). This PR can close when #1433 merges — leaving that to the owner rather than closing another author's PR.

_🤖 [Claude Code](https://claude.com/claude-code)_

nugget:
Superseded by subsequent work

Issue #1444, opened 2026-08-30
feat(companion): attested device keys — companion identity joins the signed-trust model


thane-developer:
## Problem or Opportunity

Companion authentication is a shared secret. Each account carries a list of bearer tokens in server configuration; a device proves membership in an account by presenting one, and everything else it asserts about itself — including the `client_id` that #1437 turns into durable device identity — is taken on faith. This shape has three structural weaknesses, none urgent today, all of which worsen as companions multiply and roam.

Revocation granularity is the account, not the device. A lost or retired phone forces rotating a credential shared by every sibling device on that account, and re-provisioning all of them. Device inventory (#1437) makes devices first-class and durable; their credentials remain communal.

Device identity is claimed, not proven. The durable inventory keys devices by `(account, client_id)`, but `client_id` is a string the device asserts about itself. Any holder of the account token can connect as, or submit observations attributed to, any device in that account. The inventory is honest about reachability and freshness while resting on an unverified foundation.

Bearer semantics assume a trustworthy transport boundary. Whoever observes the token is the account. The standing deployment assumption is companion connectivity over a private network (Tailscale), and under that assumption bearer tokens are adequate — but the assumption is a choice with an expiry, not a property of the system. Long-term topology goals include ingestion reachable across untrusted transit, and a long-lived replayable secret is the wrong primitive to carry into that world.

There is also an alignment argument independent of threat modeling. This repository already has one answer to "who may act": an identity is a key, and authorization is that key's attested presence in signed history — per-root trust files, operator and agent principals, sshsig verification. Companion devices are the one identity class still authenticated by a string in a config file. Extending the existing trust model to them retires a parallel credential mechanism rather than adding one.

### A companion device is a keypair

Each device generates a keypair at first run and never exports it. On iOS this is a Secure Enclave P-256 key; on macOS, Secure Enclave or Keychain equivalent. The key fingerprint becomes the device's durable identity, superseding the claimed `client_id` — which is exactly why #1437 keeps the inventory's identity column opaque.

### Enrollment is operator attestation

A device joins an account when the operator signs an enrollment record binding the key fingerprint to the account, a set of grants, and an expiry. Enrollment records are shaped like the existing per-root trust files and live in signed history, so device admission and revocation are auditable commits. Revoking a device is removing its entry — one device, not the account.

[...]

Issue #1437, opened 2026-08-29
feat(companion): persist device identity and ingest background observations


thane-developer:
## Problem or Opportunity

The companion subsystem currently equates a device with its active WebSocket connection. When `thane-agent-ios` enters the background or the phone locks, iOS closes the socket; `Registry.Remove` then removes the provider, the connected-companion context loses every reference to the phone, and its dynamically synthesized tools disappear.

Removing live tools is correct because Thane cannot dispatch an interactive request over a closed socket. Removing the device itself is not. A known operator-owned companion, its current reachability, and its last reported data are different states with different lifetimes:

1. **Known companion** — durable device identity and metadata.
2. **Live connection** — ephemeral transport over which Thane can issue interactive requests.
3. **Latest observations** — durable, bounded snapshots pushed by a companion, including their age and provenance.

Basic iOS background operation should be event-driven. When iOS wakes the app for an approved system event such as a significant location change, the companion should be able to push an observation over authenticated HTTPS without establishing or maintaining a background WebSocket. Thane should preserve the device reference and expose the latest approved observation honestly while the phone is offline.

### Persist companion device inventory

Add a companion store backed by the existing primary Thane SQLite database. Persist devices by the server-assigned account and the companion's stable `client_id`, not by the per-connection `provider_id`.

A device record should include at least:

- account and stable client ID;
- client display name and explicit platform;
- app and OS versions when supplied;
- first seen, last seen, last connected, and last disconnected times;
- the most recently advertised capability/observation manifest and its version; and
- an explicit forgotten or retired state for future lifecycle management.

Authentication and capability registration should upsert the durable record. Disconnect should update its timestamps but must not delete it. Reachability must remain derived from the in-memory registry rather than persisted as truth.

### Store latest observations

Add a latest-observation store keyed by account, client ID, and observation kind. The initial contract should retain one current record per kind, not an unbounded location history.

Each stored observation should retain:

- a unique event ID for idempotency;
- observation kind and schema version;
- `observed_at` supplied by the companion;
- authoritative server `received_at`;
- a bounded, validated payload; and
- withdrawn/unavailable state when sharing permission is revoked.

The model-facing contract must distinguish observation time from receipt time and must never describe cached location as current. Freshness/expiry semantics should be explicit and tested.

### Add authenticated observation ingestion

Add a canonical native endpoint, tentatively:

```http
POST /v1/companion/observations
Authorization: Bearer <companion token>
```

It should accept a bounded batch such as:

[...]

PR #1445, opened 2026-08-30
feat(companion): durable device inventory survives the connection

thane-developer:
First slice of #1437, per the sequencing posted there: the durable device inventory. A companion device and its WebSocket connection have different lifetimes, and until now the server conflated them — `Registry.Remove` erased every trace of a device the moment iOS backgrounded the app or a Mac went to sleep. This PR gives the first two of the issue's three states (known companion, live connection) their separate lifetimes; latest observations arrive in the next slice.

## What ships

**`internal/state/companions`** persists devices in the primary Thane database under an immutable server-assigned `device_id` primary key, minted once when a device is first seen. `(account, client_id)` — the authenticated account plus the stable identity the companion claims about itself, never the per-connection `provider_id` — is the UNIQUE lookup that maps a claim to its device. That separation is what makes the #1444 seam real: the enrollment arc makes key replacement and re-enrollment normal, so rotating a credential re-points the mapping while everything that must survive rotation (future observation rows included) references `device_id` and keeps its continuity. The schema is the standard declarative idempotent `database.Schema`; the store borrows the memory store's `thane.db` handle like its siblings.

**Lifecycle recording** hangs off the existing connection path through a new `DeviceRecorder` seam on the WebSocket handler, but never runs on it: writes are enqueued to a single ordered recording goroutine with a bounded queue, a per-write timeout, and an explicit shutdown drain registered on the app's LIFO closer stack ahead of the database owner — a slow database cannot delay `auth_ok` or teardown, and a restart cannot lose a just-recorded event or write into a closed handle. Authentication upserts the record — `first_seen_at` anchored exactly once, metadata non-empty-wins so a client that omits a field on one connection cannot erase what it reported before. Capability registration stores the most recently advertised manifest (the normalized snapshot, verbatim JSON — what the device *offered when last heard from*, not what is callable now), stamped with `capabilities_recorded_at`. Disconnect stamps `last_disconnected_at` and advances `last_seen_at` — the connection was live evidence of the device until it tore down — and never deletes or degrades the record. Because a reconnecting phone can race its own dying socket, every write is monotonic: timestamps never regress, an older racing write can only fill still-empty metadata fields, and a stale capability manifest is dropped in favor of the newer one. Recording is best-effort by design; a persistence failure logs a warning and never disturbs the live connection.

**The auth handshake grows optional device metadata** — `platform`, `app_version`, `os_version` — carried onto the provider snapshot and into the inventory. Additive and backward-compatible: clients that omit them lose nothing at the protocol level.

**The anonymous-client policy** published on #1437 is now enforced and pinned by test: a connection without a `client_id` stays fully functional live but is not inventoried. No durable identity claim, no durable row.

## What deliberately does not ship

Reachability is never persisted — `online` stays derived from the in-memory registry, exactly as the issue requires. The `state` column exists (default `active`) so a future forget/retire flow has somewhere to record its decision, but nothing writes any other value yet. No model-facing surface changes: the joined context view and reachability-semantics fix are slice 3, the last-known-location tool is slice 4.

## Verification

- [x] Known companion identity survives a process restart (`TestSurvivesReopen`: record, close, reopen, second migration idempotent, metadata and timestamps intact)
- [x] Disconnect records state without deleting it (`TestRecordDisconnectedPreservesRecord`, `TestDeviceRecorderFullLifecycle`)
- [x] Identity is keyed by account + stable `client_id`; `provider_id` never touches the store
- [x] Reconnect preserves `first_seen_at` and previously reported metadata (`TestRecordConnectedPreservesFirstSeenAndMetadata`)
- [x] Anonymous clients (no `client_id`) are never inventoried (`TestDeviceRecorderSkipsAnonymousClients`)
- [x] `just ci` passes locally (fmt, tidy, config-generate, architecture, links, lint, openapi lint, full test suite)

Refs #1437

### DIFF
--- a/internal/integrations/companion/devices.go
+++ b/internal/integrations/companion/devices.go
@@ -51,14 +51,34 @@ type DeviceRecorder interface {
 // connection lifecycle and starts the recording goroutine. Optional: a
 // nil recorder (the default) disables persistence and the handler
 // behaves as a pure live-connection endpoint. Call before the handler
-// starts serving; the queue goroutine lives for the process lifetime.
+// starts serving, and pair with [Handler.CloseDeviceRecorder] at
+// shutdown so queued writes land before their database closes.
 func (h *Handler) SetDeviceRecorder(recorder DeviceRecorder) {
 	h.devices = recorder
 	if recorder == nil || h.deviceOps != nil {
 		return
 	}
 	h.deviceOps = make(chan func(context.Context), deviceOpQueueSize)
+	h.deviceOpsDone = make(chan struct{})
 	go h.runDeviceOps()
+}
+
+// CloseDeviceRecorder stops accepting new inventory writes, drains the
+// ones already queued, and returns once the recording goroutine has
+// exited. Registered on the app's LIFO closer stack ahead of the
+// database owner, so a restart cannot lose a just-recorded lifecycle
+// event or write into a closed database. Safe to call more than once;
+// lifecycle events after close are dropped (best-effort, as ever).
+func (h *Handler) CloseDeviceRecorder() {
+	h.deviceOpsMu.Lock()
+	if h.deviceOps == nil || h.deviceOpsClosed {
+		h.deviceOpsMu.Unlock()
+		return
+	}
+	h.deviceOpsClosed = true
+	close(h.deviceOps)
+	h.deviceOpsMu.Unlock()
+	<-h.deviceOpsDone
 }
 
 // runDeviceOps applies queued inventory writes in order. A single
@@ -66,6 +86,7 @@ func (h *Handler) SetDeviceRecorder(recorder DeviceRecorder) {
 // sequence ordered without ever putting a database write on the
 // WebSocket handshake or teardown path.
 func (h *Handler) runDeviceOps() {
+	defer close(h.deviceOpsDone)
 	for op := range h.deviceOps {
 		ctx, cancel := context.WithTimeout(context.Background(), deviceOpTimeout)
 		op(ctx)
@@ -74,9 +95,15 @@ func (h *Handler) runDeviceOps() {
 }
 
 // enqueueDeviceOp hands a write to the recording goroutine without
-// blocking the connection path. Dropping on overflow is deliberate:
-// inventory is best-effort relative to live traffic.
+// blocking the connection path. Dropping on overflow or after close is
+// deliberate: inventory is best-effort relative to live traffic.
 func (h *Handler) enqueueDeviceOp(op func(context.Context)) {
+	h.deviceOpsMu.Lock()
+	defer h.deviceOpsMu.Unlock()
+	if h.deviceOpsClosed {
+		h.logger.Debug("companion device inventory recorder closed; dropping record")
+		return
+	}
 	select {
 	case h.deviceOps <- op:
 	default:
--- a/internal/integrations/companion/devices.go
+++ b/internal/integrations/companion/devices.go
@@ -1,0 +1,115 @@
+package companion
+
+import (
+	"encoding/json"
+	"strings"
+	"time"
+)
+
+// DeviceMetadata is the client-reported device description carried from
+// the auth handshake into the durable inventory. Every field is
+// optional; the inventory treats absent fields as "not reported", not
+// as an erasure of what the device said before.
+type DeviceMetadata struct {
+	ClientName string
+	Platform   string
+	AppVersion string
+	OSVersion  string
+}
+
+// DeviceRecorder receives durable device-inventory events from the
+// connection lifecycle. Connections are ephemeral and records are not:
+// a recorder must treat disconnects as timestamp updates, never as
+// deletions (#1437). The handler calls it best-effort — a recording
+// failure is logged and must not affect the live connection.
+type DeviceRecorder interface {
+	// RecordConnected upserts the device row for a successful
+	// authentication, stamping first-seen on new devices.
+	RecordConnected(account, clientID string, meta DeviceMetadata, at time.Time) error
+	// RecordCapabilities stores the most recently advertised capability
+	// manifest (opaque JSON) for the device.
+	RecordCapabilities(account, clientID string, manifest []byte, at time.Time) error
+	// RecordDisconnected stamps the disconnect time for the device.
+	RecordDisconnected(account, clientID string, at time.Time) error
+}
+
+// SetDeviceRecorder wires durable device-inventory recording into the
+// connection lifecycle. Optional: a nil recorder (the default) disables
+// persistence and the handler behaves as a pure live-connection
+// endpoint.
+func (h *Handler) SetDeviceRecorder(recorder DeviceRecorder) {
+	h.devices = recorder
+}
+
+// recordConnected persists the auth-time device upsert. Connections
+// without a client_id have no durable identity to key a row by; they
+// stay fully functional live but are deliberately not inventoried —
+// the contract published on #1437.
+func (h *Handler) recordConnected(p *Provider) {
+	if h.devices == nil {
+		return
+	}
+	if strings.TrimSpace(p.ClientID) == "" {
+		h.logger.Debug("companion connection has no client_id; not recorded in device inventory",
+			"provider_id", p.ID,
+			"account", p.Account,
+			"client_name", p.ClientName,
+		)
+		return
+	}
+	err := h.devices.RecordConnected(p.Account, p.ClientID, DeviceMetadata{
+		ClientName: p.ClientName,
+		Platform:   p.Platform,
+		AppVersion: p.AppVersion,
+		OSVersion:  p.OSVersion,
+	}, time.Now().UTC())
+	if err != nil {
+		h.logger.Warn("companion device inventory connect record failed",
+			"provider_id", p.ID,
+			"account", p.Account,
+			"client_id", p.ClientID,
+			"error", err,
+		)
+	}
+}
+
+// recordCapabilities persists the normalized capability manifest the
+// provider just registered.
+func (h *Handler) recordCapabilities(p *Provider) {
+	if h.devices == nil || strings.TrimSpace(p.ClientID) == "" {
+		return
+	}
+	manifest, err := json.Marshal(p.capabilitiesSnapshot())
+	if err != nil {
+		h.logger.Warn("companion device inventory manifest marshal failed",
+			"provider_id", p.ID,
+			"account", p.Account,
+			"client_id", p.ClientID,
+			"error", err,
+		)
+		return
+	}
+	if err := h.devices.RecordCapabilities(p.Account, p.ClientID, manifest, time.Now().UTC()); err != nil {
+		h.logger.Warn("companion device inventory capability record failed",
+			"provider_id", p.ID,
+			"account", p.Account,
+			"client_id", p.ClientID,
+			"error", err,
+		)
+	}
+}
+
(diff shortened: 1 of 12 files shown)


internal/state/companions/store.go
```diff
@@ -0,0 +1,256 @@
...
+			account, client_id, client_name, platform, app_version, os_version,
+			first_seen_at, last_seen_at, last_connected_at, state
+		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+		ON CONFLICT(account, client_id) DO UPDATE SET
+			client_name       = CASE WHEN excluded.client_name != '' THEN excluded.client_name ELSE client_name END,
+			platform          = CASE WHEN excluded.platform != '' THEN excluded.platform ELSE platform END,
+			app_version       = CASE WHEN excluded.app_version != '' THEN excluded.app_version ELSE app_version END,
+			os_version        = CASE WHEN excluded.os_version != '' THEN excluded.os_version ELSE os_version END,
+			last_seen_at      = excluded.last_seen_at,
+			last_connected_at = excluded.last_connected_at
```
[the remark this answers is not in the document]
thane-developer: Fixed in a26cada4 — connect timestamps are now MAX-guarded and metadata replacement is gated on the connect being at least as new (older writes only fill still-empty fields); pinned by TestOutOfOrderConnectWrites.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/state/companions/store.go
```diff
@@ -0,0 +1,256 @@
...
+	if len(manifest) == 0 {
+		manifest = []byte("[]")
+	}
+	if !json.Valid(manifest) {
+		return fmt.Errorf("record companion capabilities %s/%s: manifest is not valid JSON", account, clientID)
+	}
+
+	res, err := s.db.Exec(`
+		UPDATE companion_devices
+		SET capabilities = ?, last_seen_at = ?
```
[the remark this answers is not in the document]
thane-developer: Fixed in a26cada4 — added a capabilities_recorded_at stamp and the update is conditioned on it, so a stale manifest is dropped silently in favor of the newer one; pinned by TestStaleCapabilitiesIgnored.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/state/companions/store.go
```diff
@@ -0,0 +1,256 @@
...
+func (s *Store) RecordDisconnected(account, clientID string, at time.Time) error {
+	account, clientID, err := normalizeKey(account, clientID)
+	if err != nil {
+		return err
+	}
+	at = normalizeTime(at)
+
+	res, err := s.db.Exec(`
+		UPDATE companion_devices
+		SET last_disconnected_at = ?, last_seen_at = ?
```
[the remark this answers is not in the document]
thane-developer: Partially in a26cada4: both stamps are now monotonic for overlapping teardowns, and the PR description was corrected. The last_seen_at advance is kept deliberately — nothing else advances it mid-connection, so freezing it at connect time would misreport a long-lived session far worse than the bounded heartbeat skew; the Godoc now carries that reasoning.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/state/companions/store.go
```diff
@@ -0,0 +1,256 @@
...
+	account = strings.TrimSpace(account)
+	clientID = strings.TrimSpace(clientID)
+	if account == "" {
+		return "", "", fmt.Errorf("account is required")
+	}
+	if clientID == "" {
+		return "", "", fmt.Errorf("client_id is required")
+	}
+	return account, clientID, nil
+}
```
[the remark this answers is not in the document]
thane-developer: Fixed in a26cada4 — normalization happens exactly once, at authentication (handler trims what the client sent); the store validates non-empty but stores identity bytes verbatim, so durable and live values always match; pinned by TestClientIDStoredVerbatim.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/integrations/companion/handler.go
```diff
@@ -119,10 +120,15 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Register the provider before confirming to the client, so the
-	// registry is consistent by the time the client reads auth_ok.
+	// registry is consistent by the time the client reads auth_ok. The
+	// durable inventory upsert rides the same moment; the disconnect
+	// stamp mirrors it on the way out — timestamps only, the record
+	// itself outlives the connection (#1437).
 	h.registry.Add(provider)
+	h.recordConnected(provider)
```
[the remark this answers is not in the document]
thane-developer: Fixed in a26cada4 — inventory writes moved off the connection path onto a single ordered recording goroutine with a bounded queue and a 5s per-write context; timestamps are captured at event time so the monotonic guards stay meaningful.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/integrations/companion/devices_test.go
```diff
@@ -0,0 +1,240 @@
...
+	var result Message
+	readJSON(t, conn, &result)
+	if !result.Success {
+		t.Fatalf("capability registration failed: %+v", result)
+	}
+	conn.Close()
+
+	// The disconnect path is the last recorder call site; give the
+	// teardown goroutine time to run before asserting silence.
+	time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
thane-developer: Fixed in a26cada4 — the sleep is gone; the test wraps ServeHTTP to signal teardown completion, then pushes a barrier op through the recording queue before asserting the recorder stayed silent.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/state/companions/schema.go
```diff
@@ -0,0 +1,39 @@
...
+				app_version          TEXT NOT NULL DEFAULT '',
+				os_version           TEXT NOT NULL DEFAULT '',
+				first_seen_at        TIMESTAMP NOT NULL,
+				last_seen_at         TIMESTAMP NOT NULL,
+				last_connected_at    TIMESTAMP,
+				last_disconnected_at TIMESTAMP,
+				capabilities         TEXT NOT NULL DEFAULT '[]',
+				capabilities_recorded_at TIMESTAMP,
+				state                TEXT NOT NULL DEFAULT 'active',
+				PRIMARY KEY (account, client_id)
```
thane-developer: The planned credential rotation would replace the durable device at this boundary. #1444 says the key fingerprint eventually supersedes client_id, but it also makes Secure Enclave re-enrollment and key replacement normal. Once slice 2 references this composite key from observations, rotating a key either creates a second device or requires rewriting every dependent row. Please give companion_devices an immutable server-side device_id primary key now, retain UNIQUE(account, client_id) as the legacy lookup, and let the future credential table map fingerprints to device_id. That preserves inventory and observation continuity across re-enrollment and makes the claimed no-schema-break seam real.

internal/integrations/companion/devices.go
```diff
@@ -0,0 +1,169 @@
...
+// nil recorder (the default) disables persistence and the handler
+// behaves as a pure live-connection endpoint. Call before the handler
+// starts serving; the queue goroutine lives for the process lifetime.
+func (h *Handler) SetDeviceRecorder(recorder DeviceRecorder) {
+	h.devices = recorder
+	if recorder == nil || h.deviceOps != nil {
+		return
+	}
+	h.deviceOps = make(chan func(context.Context), deviceOpQueueSize)
+	go h.runDeviceOps()
```
thane-developer, line 63: This queue has no shutdown or drain path. App shutdown eventually closes the borrowed memory database, so a recently queued connect or capability write can be lost during a normal restart or can run after the DB is closed; TestSurvivesReopen exercises the synchronous store rather than this production path. Please add an explicit close/drain lifecycle tied into the App closer stack before memory closes, safely stop accepting new operations, and wait for queued writes. That also prevents one blocked goroutine from leaking from every handler-backed test.

internal/state/companions/store.go
```diff
@@ -0,0 +1,307 @@
...
+		return err
+	}
+	at = normalizeTime(at)
+
+	_, err := s.db.ExecContext(ctx, `
+		INSERT INTO companion_devices (
+			account, client_id, client_name, platform, app_version, os_version,
+			first_seen_at, last_seen_at, last_connected_at, state
+		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+		ON CONFLICT(account, client_id) DO UPDATE SET
```
thane-developer, line 118: The asynchronous queue means row-creation order is not necessarily event-time order: connection A can capture t0, be descheduled before enqueue, and connection B can capture/enqueue t1 first. The later-arriving t0 write then leaves first_seen_at at t1 even though the server observed the device at t0; the new test currently codifies that scheduler-dependent result while every other timestamp uses event-time monotonicity. Please set first_seen_at = MIN(companion_devices.first_seen_at, excluded.first_seen_at) in the conflict update and assert t0 in the out-of-order test.

thane-developer (commented):
Three net-new contract and lifecycle concerns after rechecking the updated head. The first-round fixes and focused race tests look clean.

internal/state/companions/schema.go
thane-developer: Fixed in a81d961b — companion_devices now has an immutable server-minted device_id PRIMARY KEY with UNIQUE(account, client_id) as the claim lookup, so credential rotation re-points the mapping and slice 2's observations will reference device_id; pinned by TestDeviceIDStableAcrossReconnects. Good catch — the opaque column alone wasn't a real seam once re-enrollment is normal.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/integrations/companion/devices.go
thane-developer: Fixed in a81d961b — CloseDeviceRecorder stops intake, drains the queue, and waits for the goroutine to exit; it registers on the app's LIFO closer stack after the memory store's closer so it runs before the database closes. Handler-backed tests now tear the goroutine down via t.Cleanup; pinned by TestCloseDeviceRecorderDrains.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/state/companions/store.go
thane-developer: Fixed in a81d961b — first_seen_at = MIN(existing, excluded) in the conflict update, and TestOutOfOrderConnectWrites now asserts the earlier event time instead of codifying the scheduler-dependent result.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/state/companions/schema.go
thane-developer: Verified in a81d961b: device_id is now the immutable primary key, the current (account, client_id) claim is only a unique lookup mapping, and the reconnect test pins identity stability. This addresses my concern. Thanks.

internal/state/companions/store.go
thane-developer: Verified in a81d961b: the conflict update now MIN-guards first_seen_at, and the out-of-order test asserts the earlier event time. This addresses my concern. Thanks.

internal/integrations/companion/devices.go
thane-developer: Verified in a81d961b: enqueue and close are serialized, close drains and waits for the recorder goroutine, and the LIFO app closer runs before the memory DB closer. The focused race suite and full local just ci both pass. This addresses my concern. Thanks.

nugget (approved):

PR #1446, opened 2026-08-30
feat(companion): ingest latest background observations

thane-developer:
## Summary

- add authenticated, bounded `POST /v1/companion/observations` ingestion on top of the device inventory merged in #1445
- resolve the bearer account plus opaque `client_id` claim to the inventory’s immutable server-assigned `device_id` before storage
- retain one latest value per device and observation kind, with idempotent event IDs, independent observation/receipt times, privacy-safe withdrawal ordering, and a 64-kind per-device bound
- refresh non-empty durable device metadata from successful background uploads using server receipt time as the cross-transport recency guard, including a migration backfill for existing #1445 rows
- document the private-network bearer assumption, RFC 9421-shaped authentication seam, limits, errors, and native API contract

## Why

iOS suspends its realtime WebSocket when the app backgrounds or the phone locks. A significant-change Core Location wake needs a finite authenticated HTTP path that does not depend on a live provider connection.

This is slice 2 of #1437 and the critical server contract for `thane-agent-ios`. It builds on #1445 instead of duplicating inventory. Current bearer tokens are acceptable only within the explicit private-network/Tailscale deployment boundary. The authenticator receives the bounded method, scheme, authority, target URI, headers, exact body, and claimed client identity, then returns a resolved account plus immutable device identity. That leaves ingestion unchanged when #1444 introduces RFC 9421 device signatures. Because production terminates TLS at a reverse proxy, #1444 must also configure a trusted external scheme/origin before treating `@target-uri` as authoritative; this slice deliberately does not trust client-supplied forwarding headers.

## User impact

A previously registered iOS companion can submit queued `ios.location` and `ios.system-context` events during a short background execution window, even though it has no live WebSocket. Successful uploads advance durable last-seen and non-empty device metadata without making the device remotely callable.

The endpoint rejects implausibly old clocks, bounds top-level payload properties and durable kind cardinality, and makes a same-time withdrawal dominate availability so sensitive payloads cannot survive or be resurrected by a tie-break. When companion auth is disabled, the documented route remains present and returns a structured 503.

This slice stores the data without exposing it ambiently. The joined offline context and `ios_last_known_location` tool remain separate follow-on slices, as requested in the #1437 implementation plan.

## Test plan

- `just ci`
- `go test -race ./internal/integrations/companion ./internal/state/companions`
- repeated concurrency/order stress coverage for writer serialization, replay immutability, equal-time withdrawal dominance, tombstone recovery, and at-cap kind updates
- legacy #1445 schema migration coverage for the additive column, metadata-recency backfill, stale-write rejection, and post-upgrade ingestion
- authenticated handler coverage for exact bounded RFC 9421 request material, unknown devices, unavailable auth, malformed input, body/event/payload limits, inclusive lower/maximum bounds, and withdrawals
- store coverage for immutable device-ID resolution, account isolation, cross-transport metadata recency, replay/out-of-order handling, last-seen advancement, routing-scope errors, ambiguity, and database restart survival
- OpenAPI lint and route/schema coverage

Refs #1437
Builds on #1445


internal/state/companions/observations.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := tx.QueryRowContext(ctx,
+		`SELECT account FROM companion_devices WHERE device_id = ?`, principal.DeviceID,
+	).Scan(&storedAccount); err != nil {
+		return result, fmt.Errorf("verify companion observation device %s: %w", principal.DeviceID, err)
+	}
+	if storedAccount != principal.Account {
+		return result, fmt.Errorf("companion observation device does not belong to authenticated account")
+	}
+	if _, err := tx.ExecContext(ctx, `
+		UPDATE companion_devices
```
thane-developer, line 68: The batch's `ObservationDeviceMetadata` (client_name, platform, app_version, os_version) is validated and trimmed by the handler, but ingestion only advances `last_seen_at` — the metadata is silently dropped. Contract surface that accepts data and does nothing with it is a documentation hazard: a client author will reasonably assume a background upload refreshes device metadata. Suggest either persisting it non-empty-wins with the same recency gate `RecordConnected` uses, or removing the optional fields from the wire contract so the batch carries only `client_id`. Fine as a fast-follow; flagging so it doesn't fossilize.

internal/integrations/companion/observation_handler.go
```diff
@@ -0,0 +1,274 @@
...
+			return err
+		}
+		if event.SchemaVersion < 1 || event.SchemaVersion > 1000 {
+			return fmt.Errorf("events[%d].schema_version must be between 1 and 1000", i)
+		}
+		if event.ObservedAt.IsZero() {
+			return fmt.Errorf("events[%d].observed_at is required", i)
+		}
+		if event.ObservedAt.After(receivedAt.Add(maxObservationFutureSkew)) {
+			return fmt.Errorf("events[%d].observed_at is too far in the future", i)
```
thane-developer, line 196: Small graft from the closed #1447 worth considering: a lower plausibility floor on `observed_at` (it rejected anything before 2000-01-01). `IsZero` catches the Go zero value but a device with an unset clock reporting 1970 sails through. It's mostly harmless in a latest-wins store — an epoch-dated row loses to any real observation — but it can become the *only* stored row for a kind and then honestly report a 56-year observation age. Terminal-invalid at the edge is cheaper than explaining that later.

thane-developer (commented):
Reviewed in full and ran `just ci` on this branch locally — it passes. This PR is adopted as the slice-2 implementation for #1437, and the parallel #1447 is closed in its favor.

The deciding factors, for the record: the authenticator seam here receives the exact bounded request material (method, target, headers, body), which is the shape RFC 9421 verification in #1444 actually needs — the competing implementation authenticated before the body was read and would have required re-architecture. The principal resolving to the immutable `device_id` with unknown-device-as-401 is stricter and mirrors how device-key auth will work. And the read-side `ResolveLatestObservation` ambiguity semantics pre-build slice 4's tool substrate in the same routing shape the live registry already uses.

Two findings inline; neither blocks merge.

internal/integrations/companion/observation_handler.go
```diff
@@ -0,0 +1,274 @@
...
+		event := &batch.Events[i]
+		event.EventID = strings.TrimSpace(event.EventID)
+		event.Kind = strings.TrimSpace(event.Kind)
+		parsedEventID, err := uuid.Parse(event.EventID)
+		if err != nil {
+			return fmt.Errorf("events[%d].event_id must be a UUID", i)
+		}
+		event.EventID = parsedEventID.String()
+		if err := validateBoundedIdentifier(fmt.Sprintf("events[%d].kind", i), event.Kind); err != nil {
+			return err
```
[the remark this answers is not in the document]
thane-developer: Implemented in 767f8b7f with an atomic per-device cardinality bound rather than a kind allowlist, preserving extensible namespaced kinds. After acquiring the SQLite writer slot, ingestion preflights existing plus new unique kinds and rejects creation beyond 64 with a structured 400; existing kinds can continue advancing. The transaction rolls back the whole batch on overflow.

internal/state/companions/observations.go
```diff
@@ -0,0 +1,218 @@
...
+				event_id = excluded.event_id,
+				schema_version = excluded.schema_version,
+				status = excluded.status,
+				observed_at = excluded.observed_at,
+				received_at = excluded.received_at,
+				payload = excluded.payload
+			WHERE excluded.event_id != companion_latest_observations.event_id
+			  AND (excluded.observed_at > companion_latest_observations.observed_at
+			       OR (excluded.observed_at = companion_latest_observations.observed_at
+			           AND excluded.event_id > companion_latest_observations.event_id))
```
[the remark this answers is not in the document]
thane-developer: Implemented in 767f8b7f. At equal observed_at values, withdrawn now dominates available regardless of event ID; an equal-time available event cannot resurrect a tombstone. Event ID remains the deterministic tie-break only when statuses match. Tests pin both directions.

internal/integrations/companion/observation_auth.go
```diff
@@ -0,0 +1,101 @@
...
+
+// ObservationAuthRequest is the bounded HTTP material an authenticator may
+// verify. Keeping the request target, headers, and exact body here allows a
+// future RFC 9421 authenticator to replace bearer auth without changing the
+// ingestion service contract.
+type ObservationAuthRequest struct {
+	Method          string
+	RequestTarget   string
+	Header          http.Header
+	Body            []byte
```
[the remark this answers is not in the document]
thane-developer: Implemented in 767f8b7f. ObservationAuthRequest now carries Scheme, Authority, RequestTarget, and the composed absolute TargetURI in addition to method, cloned headers, and exact bounded body. The handler derives HTTPS from TLS (or an absolute request URL) and tests the full @target-uri material.

internal/app/new_servers.go
```diff
@@ -283,6 +283,11 @@ func (a *App) initServers(s *newState) error {
 		handler.SetDeviceRecorder(a.companionDevices)
 		a.onClose("companion-device-recorder", handler.CloseDeviceRecorder)
 		server.SetCompanionHandler(handler)
+		server.SetCompanionObservationHandler(companion.NewObservationHandler(
+			companion.NewBearerObservationAuthenticator(cfg.Companion.TokenIndex(), a.companionDevices.ResolveObservationIdentity),
+			a.companionDevices,
+			logger,
+		))
```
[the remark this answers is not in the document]
thane-developer: Implemented in 767f8b7f. App wiring now always installs ObservationHandler with the durable store; when companion config is disabled it receives a nil authenticator and returns the documented JSON 503 observation_auth_unavailable. WebSocket/registry setup remains conditional.

internal/state/companions/observations.go
```diff
@@ -0,0 +1,218 @@
...
+	if storedAccount != principal.Account {
+		return result, fmt.Errorf("companion observation device does not belong to authenticated account")
+	}
+	if _, err := tx.ExecContext(ctx, `
+		UPDATE companion_devices
+		SET last_seen_at = MAX(last_seen_at, ?)
+		WHERE device_id = ? AND account = ?
+	`, result.ReceivedAt, principal.DeviceID, principal.Account); err != nil {
+		return result, fmt.Errorf("update companion observation last seen: %w", err)
+	}
```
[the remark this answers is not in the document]
thane-developer: Implemented in 767f8b7f. The account-scoped companion_devices UPDATE is now the first transaction statement, and RowsAffected verifies device ownership before any read. The kind preflight runs only after that writer acquisition. A file-backed WAL concurrency test plus a 20-run stress pass verifies concurrent uploads serialize cleanly.

internal/state/companions/observations.go
thane-developer: Implemented in 767f8b7f. Non-empty observation metadata now refreshes the durable device row. I added metadata_recorded_at as a shared cross-transport recency guard and moved RecordConnected onto it too, so newer WebSocket/background reports win while empty fields never erase prior values.

internal/server/openapi/native.yaml
```diff
@@ -1293,6 +1407,137 @@ components:
...
+          description: Whether payload is available or was explicitly withdrawn.
+          example: available
+        observed_at:
+          type: string
+          format: date-time
+          description: Device-supplied observation time, independent of server receipt time.
+          example: "2026-08-30T16:59:55.123Z"
+        payload:
+          type: object
+          maxProperties: 256
```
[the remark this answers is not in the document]
thane-developer: Implemented in 767f8b7f. Available payloads are now decoded as JSON objects and rejected above 256 top-level properties, matching maxProperties in OpenAPI. Focused coverage constructs a 257-property payload.

internal/integrations/companion/observation_handler.go
thane-developer: Implemented in 767f8b7f. observed_at values before 2000-01-01T00:00:00Z are now rejected at the handler edge, with focused coverage and the lower bound documented in both OpenAPI and the API reference.

internal/state/companions/store.go
```diff
@@ -112,28 +115,36 @@ func (s *Store) RecordConnected(ctx context.Context, account, clientID string, m
...
-			device_id, account, client_id, client_name, platform, app_version, os_version,
+			device_id, account, client_id, client_name, platform, app_version, os_version, metadata_recorded_at,
 			first_seen_at, last_seen_at, last_connected_at, state
-		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
 		ON CONFLICT(account, client_id) DO UPDATE SET
 			first_seen_at = MIN(companion_devices.first_seen_at, excluded.first_seen_at),
 			client_name = CASE WHEN excluded.client_name != ''
-					AND (companion_devices.client_name = '' OR excluded.last_connected_at >= companion_devices.last_connected_at)
+					AND (companion_devices.client_name = '' OR companion_devices.metadata_recorded_at IS NULL
```
thane-developer, line 124: **[medium — merge-gating]** Migrated slice-1 rows lose metadata recency protection. The `ColumnAdd` gives every existing prod row `metadata_recorded_at = NULL`, and this guard's `companion_devices.metadata_recorded_at IS NULL` disjunct then accepts ANY non-empty write regardless of event-time order — the old guard keyed on `last_connected_at` (populated on every slice-1 row) always applied, so this is strictly weaker on migrated rows. Verified empirically: an older queued write overwrites a newer non-empty `client_name` on a legacy-shaped row. Fix shape: add a `database.Raw` backfill after the ColumnAdd — `UPDATE companion_devices SET metadata_recorded_at = last_connected_at WHERE metadata_recorded_at IS NULL AND (client_name != '' OR platform != '' OR app_version != '' OR os_version != '')` — plus the legacy-shape test described in the schema.go comment.

internal/state/companions/schema.go
```diff
@@ -40,5 +41,28 @@ var devicesSchema = database.Schema{
 				UNIQUE (account, client_id)
 			)`,
 		},
+		database.ColumnAdd{
```
thane-developer, line 47: **[medium — merge-gating]** The upgrade path from the already-merged #1445 table shape is untested, and this `ColumnAdd` is the only thing standing between the next deploy and runtime `no such column: metadata_recorded_at` failures — yet it reads as pure duplication of the column already declared in the TableCreate above it, with no comment saying it is load-bearing. Every test starts from a fresh DB where TableCreate supplies the column, so CI would stay green if a cleanup pass deleted this step. Suggest: (1) a comment marking it as the existing-database upgrade path (sibling schemas comment exactly this pattern), and (2) a test that executes the slice-1 CREATE TABLE SQL verbatim (from 8ba2559a), seeds a row, runs `NewStore`, and asserts `RecordConnected` + `IngestObservations` succeed. Verified the upgrade works *today* — the finding is the missing pin, not the code.

internal/integrations/companion/observation_handler.go
```diff
@@ -0,0 +1,303 @@
...
+			event.Payload = nil
+		default:
+			return fmt.Errorf("events[%d].status must be available or withdrawn", i)
+		}
+		event.ObservedAt = event.ObservedAt.UTC()
+	}
+	return nil
+}
+
+func observationRequestScheme(r *http.Request) string {
```
thane-developer, line 229: **[medium — deferrable to #1444, but document now]** `observationRequestScheme` derives the scheme from `r.URL.Scheme`/`r.TLS` and never consults `X-Forwarded-Proto`. Production terminates TLS at the reverse proxy, so the Go server sees `r.TLS == nil` and this seam will hand a future RFC 9421 authenticator `Scheme: "http"` and a `TargetURI` beginning `http://` while every companion signs over `https://...` — universal verification failure. The only test uses `httptest.NewRequest` with an absolute URL, which populates the branches real origin-form requests never take. Blindly trusting `X-Forwarded-Proto` would let a direct-connect client forge the scheme, so this needs a deliberate trusted-proxy decision — fine to make in #1444, but please leave a comment here recording the caveat so the seam's 9421-readiness claim doesn't oversell.

internal/integrations/companion/observation_auth.go
```diff
@@ -0,0 +1,105 @@
...
+			account = credential.account
+		}
+	}
+	if matched != 1 {
+		return ObservationPrincipal{}, ErrObservationUnauthorized
+	}
+	if a.identityLookup == nil {
+		return ObservationPrincipal{}, fmt.Errorf("resolve companion observation identity: resolver unavailable")
+	}
+	deviceID, found, err := a.identityLookup(ctx, account, claimedClientID)
```
thane-developer, line 102: **[low]** Timing oracle: the identity lookup (a SQLite round trip) runs only after a bearer digest matches, so invalid-token and valid-token-unknown-device 401s — byte-identical on the wire — differ by up to the 5s busy_timeout in latency. Within the Tailscale boundary this is acceptable risk, and the WS handshake has the same property; flagging so the acceptance is deliberate. The cheap equalizer, if wanted: run the lookup unconditionally with a dummy account on non-match.

internal/server/openapi/native.yaml
```diff
@@ -1293,6 +1409,137 @@ components:
...
+
+    CompanionObservationIngestResult:
+      type: object
+      description: Counts showing how the accepted batch affected latest-value storage.
+      required: [stored, ignored, received_at]
+      properties:
+        stored:
+          type: integer
+          minimum: 0
+          description: Number of event kinds whose latest stored value advanced.
```
thane-developer: **[low]** `stored` is described as "Number of event kinds whose latest stored value advanced", but the implementation increments it once per applied *event* (two advancing events for one kind in a batch → `stored: 2`), and api.md documents per-event semantics. The spec should say events, matching code and api.md.

internal/server/openapi/native.yaml
```diff
@@ -1293,6 +1409,137 @@ components:
...
+        client_name:
+          type: string
+          maxLength: 128
+          description: Operator-readable device name supplied by the companion and used to refresh durable inventory metadata.
+          example: Thane for iOS
+        platform:
+          type: string
+          maxLength: 128
+          pattern: '^[A-Za-z0-9._-]+$'
+          description: Lowercase operating-system family reported by the companion and used to refresh durable inventory metadata.
```
thane-developer: **[low]** `platform`'s description says "Lowercase operating-system family" but the sibling pattern (`^[A-Za-z0-9._-]+$`) and the handler both accept uppercase, and the value is stored verbatim (per the slice-1 no-normalization invariant). Either drop "Lowercase" from the description or enforce it — as written, `"iOS"` and `"ios"` silently become two platform values while the spec implies they can't.

internal/state/companions/observations.go
```diff
@@ -0,0 +1,282 @@
...
+				observed_at, received_at, payload
+			) VALUES (?, ?, ?, ?, ?, ?, ?, ?)
+			ON CONFLICT(device_id, kind) DO UPDATE SET
+				event_id = excluded.event_id,
+				schema_version = excluded.schema_version,
+				status = excluded.status,
+				observed_at = excluded.observed_at,
+				received_at = excluded.received_at,
+				payload = excluded.payload
+			WHERE excluded.event_id != companion_latest_observations.event_id
```
thane-developer, line 158: **[low — mutation-verified]** Deleting the `excluded.event_id != companion_latest_observations.event_id` guard passes the full suite. The only replay test resends a byte-identical batch, which the observed_at ladder already ignores — so the guard's sole observable effect (blocking a same-event_id resend carrying a *newer* observed_at or different payload, e.g. after a device clock correction) is unpinned. One test: same event_id, observed_at +2s, different payload → must be Ignored.

internal/state/companions/observations.go
```diff
@@ -0,0 +1,282 @@
...
+		return result, fmt.Errorf("close companion observation kinds: %w", err)
+	}
+	newKinds := make(map[string]struct{}, len(batch.Events))
+	for _, event := range batch.Events {
+		if _, exists := existingKinds[event.Kind]; exists {
+			continue
+		}
+		newKinds[event.Kind] = struct{}{}
+	}
+	if len(newKinds) > 0 && len(existingKinds)+len(newKinds) > companion.MaxObservationKindsPerDevice {
```
thane-developer, line 132: **[low — mutation-verified]** Replacing the cap guard with one that counts batch events instead of genuinely-new kinds passes the full suite — meaning "existing kinds may continue to advance after the limit is reached" is unpinned, and a regression here would permanently brick every at-cap device. One test: seed a device to the cap, then ingest a batch of only-existing kinds and assert success.

internal/state/companions/store.go
```diff
@@ -303,6 +318,13 @@ func (s *Store) scanDevices(ctx context.Context, query string, args ...any) ([]D
 	return devices, rows.Err()
 }
 
+func deviceMetadataTime(meta companion.DeviceMetadata, at time.Time) any {
```
thane-developer, line 321: **[low — mutation-verified]** Making `deviceMetadataTime` unconditionally return the connect time (i.e. stamping `metadata_recorded_at` on metadata-free connects) passes the full suite, and no test interleaves WS `RecordConnected` with HTTP `IngestObservations` out of order. The failure that unpins: a metadata-free WS connect at T2 stamps the guard, and a queued HTTP batch carrying the real metadata at T1 < T2 is then judged stale forever. Two tests close it: metadata-free connect leaves the stamp untouched; WS→HTTP→WS out-of-order metadata converges on the newest non-empty values.

internal/integrations/companion/observation_handler_test.go
```diff
@@ -0,0 +1,265 @@
...
+	tooManyPropertiesJSON, err := json.Marshal(tooManyProperties)
+	if err != nil {
+		t.Fatalf("marshal oversized property set: %v", err)
+	}
+	tests := []struct {
+		name  string
+		batch ObservationBatch
+	}{
+		{name: "empty events", batch: ObservationBatch{ObservationDeviceMetadata: ObservationDeviceMetadata{ClientID: "iphone-1"}}},
+		{name: "too many events", batch: ObservationBatch{ObservationDeviceMetadata: ObservationDeviceMetadata{ClientID: "iphone-1"}, Events: make([]ObservationEvent, maxObservationEvents+1)}},
```
thane-developer: **[low]** The "too many events" case builds 17 zero-valued events, each of which fails `event_id` validation on its own — so the case passes even with the count check deleted (mutation-verified). Build the oversize fixture from 17 individually *valid* events, or assert the specific error, so the `maxObservationEvents` bound is actually pinned.

internal/state/companions/observations_test.go
```diff
@@ -0,0 +1,294 @@
...
+	if err != nil || !found || !updated.LastSeenAt.Equal(withdrawnAt.Add(time.Second)) {
+		t.Fatalf("last seen = %v found=%t err=%v", updated.LastSeenAt, found, err)
+	}
+	if updated.ClientName != "Nugget's iPhone" || updated.AppVersion != "1.1 (2)" || updated.OSVersion != "26.6" ||
+		!updated.MetadataRecordedAt.Equal(withdrawnAt.Add(time.Second)) {
+		t.Fatalf("observation metadata was not persisted: %+v", updated)
+	}
+}
+
+func TestIngestObservationsEqualTimeWithdrawalDominates(t *testing.T) {
```
thane-developer, line 120: **[low]** The tombstone-recovery cell of the tie-break matrix — a strictly *newer* available observation superseding a stored withdrawal — has no test. It's the only path a device ever recovers from a tombstone, it lives in the status-agnostic first disjunct of the upsert WHERE clause, and it's directly adjacent to the equal-time dominance boundary the last review round reshaped. Verified correct today by probe; one test pins it.

internal/integrations/companion/observation_handler.go
```diff
@@ -0,0 +1,303 @@
...
+		if err := validateBoundedIdentifier(fmt.Sprintf("events[%d].kind", i), event.Kind); err != nil {
+			return err
+		}
+		if event.SchemaVersion < 1 || event.SchemaVersion > 1000 {
+			return fmt.Errorf("events[%d].schema_version must be between 1 and 1000", i)
+		}
+		if event.ObservedAt.IsZero() {
+			return fmt.Errorf("events[%d].observed_at is required", i)
+		}
+		if event.ObservedAt.Before(minimumObservationTime) {
```
thane-developer, line 192: **[low]** Only the reject sides of two boundaries are tested. `observed_at` exactly `2000-01-01T00:00:00Z` (documented "on or after") and a payload with exactly 256 properties (documented `maxProperties: 256`) have no accepting test, so an off-by-one refactor (`Before` → `!After`, `>` → `>=`) would reject documented-valid input with the suite green.

thane-developer (commented):
Second full pass at 767f8b7f, run as a multi-agent review with adversarial verification: six reviewers attacked separate dimensions (transaction/tie-break correctness, schema migration, auth/security, wire-contract parity, slice-1 regression, test adequacy), and every candidate finding then had to survive three independent refutation attempts, several by actually mutating the code and re-running the suite. `just ci` verified green locally on this branch.

First, what held up: **all eight prior-round fixes do what they claim.** The write-first transaction ordering was empirically stress-tested at ~40x the scale of the original concern and never produced `SQLITE_BUSY_SNAPSHOT` — that finding is refuted and closed. The equal-time withdrawn-dominance SQL is correct across every cell of the tie-break matrix, and the strictly-newer recovery path works (verified by probe test).

12 findings survived verification — 3 medium, 9 low. Two of the mediums are, in my view, worth fixing **before merge** because they are prod-upgrade correctness on the already-merged #1445 table:

1. **NULL `metadata_recorded_at` on migrated rows weakens the slice-1 metadata guard** (inline at store.go) — needs a backfill step.
2. **The legacy-shape upgrade path is completely unpinned** (inline at schema.go) — the load-bearing `ColumnAdd` reads as redundant duplication and CI never exercises it.

The third medium (RFC 9421 scheme derivation behind the TLS-terminating proxy) has zero effect on today's bearer path and can defer to #1444 — but it deserves a comment in the code now so the seam's readiness claim carries its known caveat.

The nine lows are mostly **unpinned invariants found by mutation testing**: in each case a guard the contract documents can be deleted or inverted and the full suite stays green. Given how many rounds of review have reshaped exactly these boundaries, pinning them is cheap insurance. Details inline.

internal/state/companions/store.go
thane-developer: Fixed in 02203857. A post-ColumnAdd idempotent backfill now anchors non-empty legacy metadata to last_connected_at, preserving the #1445 recency guard. The legacy-schema test verifies that an older queued RecordConnected write cannot overwrite the migrated row.

internal/integrations/companion/observation_auth.go
thane-developer: Accepted and documented in 02203857. I am deliberately not adding dummy SQLite work to invalid bearer requests: this temporary bearer path is confined to the private-network boundary, dummy lookups amplify unauthenticated traffic without making the full exchange constant-time, and #1444 replaces the shared-secret posture. The behavior now has an explicit code comment rather than being accidental.

internal/integrations/companion/observation_handler.go
thane-developer: Documented in 02203857 and the refreshed PR body. The handler intentionally does not trust X-Forwarded-Proto from direct clients; #1444 must supply an explicitly trusted external proxy/origin scheme before treating @target-uri as authoritative under production TLS termination.

internal/state/companions/schema.go
thane-developer: Fixed in 02203857. The ColumnAdd now has an explicit load-bearing upgrade comment. TestMigratesSliceOneDeviceSchema creates the exact #1445 table shape from 8ba2559a, seeds a row, runs NewStore, verifies the backfill/stale-write guard, and successfully ingests a background observation afterward.

internal/server/openapi/native.yaml
```diff
@@ -1293,6 +1409,137 @@ components:
...
+          example: ios
```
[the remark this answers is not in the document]
thane-developer: Fixed in 02203857. The OpenAPI description now says operating-system family without claiming lowercase normalization, matching both the accepted identifier pattern and verbatim persistence.

internal/state/companions/observations_test.go
thane-developer: Pinned in 02203857. The withdrawal test now proves a strictly newer available observation advances the tombstone back to available with its payload restored, while the equal-time resurrection remains ignored.

internal/state/companions/store.go
thane-developer: Pinned in 02203857. Cross-transport coverage now records WebSocket metadata, advances it through HTTP, rejects a late stale WebSocket report, and verifies a newer metadata-free connect advances connection time without touching metadata_recorded_at.

internal/state/companions/observations.go
thane-developer: Pinned in 02203857. After seeding exactly 64 kinds, the test advances an existing kind successfully before exercising atomic rejection of a mixed existing/new overflow batch.

internal/server/openapi/native.yaml
thane-developer: Fixed in 02203857. The OpenAPI stored description now counts applied events, matching implementation and api.md when multiple events for one kind advance within a batch.

internal/state/companions/observations.go
thane-developer: Pinned in 02203857. The replay test now resends the same event_id with a newer observed_at and different payload, and asserts Ignored plus an unchanged stored timestamp/payload.

internal/server/openapi/native.yaml
thane-developer: Fixed in 02203857. Dropped the lowercase claim from the platform description so the contract matches the identifier pattern and the slice-1 verbatim-storage invariant.

internal/integrations/companion/observation_handler_test.go
thane-developer: Fixed in 02203857. The 17-event fixture now contains individually valid UUIDs, kinds, timestamps, versions, and payloads, so removing the batch-count guard makes the test fail.

internal/integrations/companion/observation_handler.go
thane-developer: Pinned in 02203857. A positive boundary test now accepts observed_at exactly 2000-01-01T00:00:00Z and a payload with exactly 256 top-level properties.

thane-developer:
Also addressed Copilot's suppressed not-found finding in 02203857: ResolveLatestObservation now describes the routing scope actually supplied (unscoped, account-only, client-only, or account plus client), with table-driven coverage for all four cases.

nugget (approved):

PR #1447, opened 2026-08-30
feat(companion): background observation ingestion over authenticated HTTPS

thane-developer:
Second slice of #1437, and the critical path for `thane-agent-ios`: an offline companion can now push observations without holding a WebSocket. iOS wakes the app for an approved system event, the app drains its durable outbox to `POST /v1/companion/observations`, and Thane remembers what the phone last reported — with observation time and receipt time recorded independently and never conflated. Builds directly on #1445's inventory: observation rows reference the immutable `device_id`, so they survive the credential rotation #1444 will make routine.

## The store

`companion_observations` keeps one current record per `(device, kind)` — a latest-observation store, not a location history, per the issue's deliberate retention contract. Verdicts are decided inside a transaction with latest-only semantics: the newest device-claimed `observed_at` wins, an exact `event_id` replay is a `duplicate`, and anything older (or tied with a distinct event) is `superseded`. Every verdict is terminal for the client's outbox entry, which makes the retry story trivial: a batch that hits a 5xx is retried verbatim, and idempotency absorbs it.

A withdrawal event clears the stored payload while keeping provenance (kind, times, event id) — revoked sharing must not leave precise location looking available — and a newer present observation resurrects the kind when sharing returns.

`EnsureDevice` gives background-only devices an inventory row: a phone that pushes before ever connecting a WebSocket still becomes a known companion, with MIN/MAX-guarded seen timestamps but no connection stamps. A successful upload means *recently seen*, never *live*.

## The endpoint

Ingestion is transport-independent behind two seams, exactly as promised on the issue. The `ObservationSink` interface is the storage contract; the `ObservationAuthenticator` interface resolves a bearer credential to an account — and this is deliberately the **first enforced authentication on a native route**, because unlike the rest of the surface this endpoint is designed to be reachable by roaming devices. The bearer implementation compares every configured companion token in constant time and never breaks early on a match; #1444 swaps in RFC 9421 signature verification behind the same interface without touching ingestion. The token alone determines the account; every accepted `client_id` is bound to it; the request body never gets a vote.

Limits are documented API contract, carried identically in code and `native.yaml`: body ≤ 256 KiB, ≤ 64 events per batch, ≤ 16 KiB per payload. Events validate individually — a malformed timestamp, a clock claiming the future beyond 5 minutes of skew, a clock reporting 1999, a non-object payload, or a withdrawal smuggling data each earn a terminal `invalid` verdict with a reason, without failing the batch around them. `observed_at` is wire-typed as a string precisely so one bad timestamp invalidates one event rather than the whole decode. Logs carry accounts, counts, and kinds; never payloads, never tokens.

## Contract surfaces

`native.yaml` declares the route with the new `companion:ingest` scope, a new `Companion` tag, full request/response schemas with examples, and an honest note that this one operation enforces authentication today. `docs/reference/api.md` carries the row. The route-coverage bijection and the OpenAPI lint gate both hold.

## For the iOS agent

The wire contract matches the issue's sketch with these specifics: `observed_at` is RFC 3339; per-event verdicts come back as `applied` / `duplicate` / `superseded` / `invalid`, all terminal (delete the outbox entry on any of them); batch-level failures are 400 (malformed, terminal), 401 (token), 413 (too big — split the batch), and 5xx (retry the batch verbatim). Withdrawals are `"withdrawn": true` with no payload. The `client_id` must be the same stable value used for WebSocket auth.

## Verification

- [x] An authenticated offline companion can submit a bounded batch over HTTPS (handler tests, full flow)
- [x] Replaying an accepted `event_id` is idempotent and cannot duplicate or regress state (`TestUpsertObservationLatestOnly`)
- [x] `received_at` recorded independently from `observed_at`; malformed and implausible timestamps handled explicitly (`TestObservationsPerEventValidation`)
- [x] A withdrawal prevents previously stored precise location from appearing available (`TestWithdrawalClearsPayload`)
- [x] One account cannot create, overwrite, or read another account's device data — the token determines the account, the body never does (`TestObservationsBindsAccountFromToken`)
- [x] Body, event-count, and payload limits documented and enforced (`TestObservationsBatchValidation`)
- [x] Observations survive a process restart (`TestObservationsSurviveReopen`)
- [x] Logs and errors expose no bearer tokens or precise payloads
- [x] Mux registration, `native.yaml`, and `docs/reference/api.md` describe the same route and schemas (coverage bijection + vacuum lint)
- [x] `just ci` passes locally

Refs #1437

### DIFF
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -101,6 +101,7 @@ filesystem paths, signer principals, or the contents of `.allowed_signers`.
 | `DELETE` | `/v1/checkpoints/{id}` | Delete a checkpoint. |
 | `POST` | `/v1/checkpoints/{id}/restore` | Restore from a checkpoint. |
 | `GET` | `/v1/realtime/ws` | First-party realtime WebSocket (canonical). |
+| `POST` | `/v1/companion/observations` | Background observation ingestion from companion devices (bearer-authenticated; enforced today). |
 | `GET` | `/v1/companion/ws` | Realtime WebSocket — legacy alias (deprecated; see below). |
 | `GET` | `/v1/platform/ws` | Realtime WebSocket — legacy alias (deprecated; see below). |
 
--- a/internal/app/new_servers.go
+++ b/internal/app/new_servers.go
@@ -284,6 +284,17 @@ func (a *App) initServers(s *newState) error {
 		a.onClose("companion-device-recorder", handler.CloseDeviceRecorder)
 		server.SetCompanionHandler(handler)
 
+		// Background observation ingestion (#1437): the authenticated
+		// HTTPS path for devices that cannot hold a WebSocket. Bearer
+		// auth rides behind the authenticator seam so the enrollment
+		// arc (#1444) can swap in signature verification without
+		// touching ingestion.
+		server.SetCompanionObservationsHandler(companion.NewObservationsHandler(
+			companion.NewTokenAuthenticator(cfg.Companion.TokenIndex()),
+			a.companionDevices,
+			logger,
+		))
+
 		a.connMgr.Watch(s.ctx, connwatch.WatcherConfig{
 			Name: "companion",
 			Probe: func(_ context.Context) error {
--- a/internal/integrations/companion/observations.go
+++ b/internal/integrations/companion/observations.go
@@ -1,0 +1,325 @@
+package companion
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"regexp"
+	"strings"
+	"time"
+)
+
+// Ingestion limits. Companion batches are outbox drains, not bulk
+// imports: a phone waking from a background event uploads a handful of
+// observations. The caps below are documented API contract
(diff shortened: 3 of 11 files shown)


thane-developer:
Closed in favor of #1446, which was built concurrently by the thane-agent-ios developer and is adopted as the slice-2 implementation after a full comparative review (and a local `just ci` verification of that branch).

The deciding factors: #1446's authenticator seam receives the exact bounded request material including the body, which is the shape RFC 9421 verification (#1444) actually needs — this PR authenticated before reading the body and would have required re-architecture to add content-digest verification. Its principal resolves to the immutable `device_id` with unknown devices rejected at auth, which is stricter and mirrors the coming device-key model. And since the same author builds the iOS client, adopting their server contract eliminates cross-repo contract skew on the critical path.

What this PR did better rides over as review comments on #1446: the implausibly-old `observed_at` floor, and the accepted-but-unused device metadata finding. The per-event verdict contract is deliberately not carried over — the aggregate stored/ignored response is the contract the iOS outbox is being built against, and latest-only idempotency makes batch-level retry safe without per-event granularity.

PR #1486, opened 2026-08-31
Repoint the archivist's dossier mandate at the dossiers root

thane-developer:
The non-contact dossiers left `kb` for a signed root of their own, and the seeded archivist mandate did not follow them. It still sent every entity, area, routine and theme to `kb:dossiers/`, which is a door that no longer opens. This repoints the teaching at `dossiers:`.

Two things change beyond the ref prefix. The root is **flat** — its documents sit at the top level, so the "namespace" framing that made sense inside `kb` is gone and the mandate says where the files actually are. And the colon warning gets sharper rather than dropped: it used to guard against writing a bare `dossiers/...` path where a `kb:` ref belonged, but now that the root is itself named `dossiers`, path and ref differ by a single character and the wrong one fails to resolve without saying so. The mandate names which separator is the ref and what the other one is, because that mistake got easier to make, not harder.

The second commit is the more interesting one. Production's live mandate had already been repointed by hand, and it does something the first pass here did not: it names `kb:dossiers/` as **retired** rather than quietly deleting every mention. That is better teaching, and this now matches it. A model that read the old mandate — or that opens a dossier still citing a `kb:dossiers` ref in its own Connections section, of which there are many in the migrated history — needs the old door explicitly closed. A door that simply vanishes from the prose reads as an oversight and invites a retry.

`TestArchivistUsesCanonicalContactDossiers` asserts the mandate's exact wording, so the assertion moves with the text it guards. `docs/understanding/document-roots.md` records the retirement so the operator-facing description of the boundary does not drift behind the model-facing one.

## Test plan

- [x] `just ci` green (run unpiped; architecture check clean, 0 lint issues, full suite)
- [x] `TestArchivistUsesCanonicalContactDossiers` passes against the new wording
- [x] No live `kb:dossiers` refs remain in `loops/` or `docs/` — only the explicit retirement notices
- [x] Seeded mandate now teaches the same thing as production's live `core/loops/archivist.md`
- [x] Line wrapping matches the file's existing convention (no new lines over 80)

### DIFF
--- a/docs/understanding/document-roots.md
+++ b/docs/understanding/document-roots.md
@@ -524,7 +524,8 @@ structured directory, reconciled with the existing canonical dossier, and
 republished through `contact_dossier_write`. The generic `dossiers` root
 remains the home for entities, areas, routines, and themes; it is never a
 fallback location for a contact, because that would split one person's
-history across two document roots. Before replacement, the archivist checks
+history across two document roots. The `kb:dossiers/` namespace this
+content used to occupy is retired, not a second home. Before replacement, the archivist checks
 whether `doc_read` truncated the current dossier and walks its outline and
 sections when necessary. An unreadable remainder or unavailable canonical
 writer defers the durable queue item behind ready work rather than losing it or
--- a/docs/understanding/document-roots.md
+++ b/docs/understanding/document-roots.md
@@ -521,9 +521,9 @@ should not author dossier structure through them.
 The archivist uses the same contract-owned door. Contact-shaped queue subjects
 and contact evidence discovered in closed sessions are resolved through the
 structured directory, reconciled with the existing canonical dossier, and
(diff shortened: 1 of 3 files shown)


thane-developer:
Superseded by #1485, which landed the same repointing while this was in flight — and did more. Main now teaches `dossiers:` refs with the retired `kb:dossiers/` namespace named explicitly, plus the archive-session citation rule and the contact-projection hygiene from #1484, and documents `dossiers` as a reserved root whose path comes from the workspace.

This branch is based on the pre-#1485 text, so merging it would revert those. Nothing here is worth rescuing: the only wording not already on main is a one-sentence gloss that the ref separator is a colon, and main already carries the "never a bare `dossiers/...` path" warning that covers the same hazard.

Closing rather than resolving the conflicts.

